PR #3, opened 2018-06-08
[WIP]add greenboot-script-runner

LorbusChris:
Fix unit files by adding [Install] section
Update README.md


greenboot-script-runner.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+run_required_scripts () {
+  required_greenboot_scripts=`ls /etc/greenboot.d/required/*.sh`
```
LorbusChris: what's the most robust way to gather all scripts in this dir and handle the empty-dir case?

dustymabe: this looks reasonable to me

greenboot-script-runner.sh
```diff
@@ -0,0 +1,36 @@
...
+  set +e
+  for script in $required_greenboot_scripts; do
+    bash $script
+    if [ $? -ne 0 ]; then
+      echo "FAILURE (return value '$?')" >&2 | systemd-cat -t "greenboot Script: '$script'" -p emerg
```
dustymabe: let's call this "exit code" rather than "return value". Also why redirect to STDERR (`>&2`) here ?

tests/passing-script.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+
```
dustymabe: maybe add an `echo "This is a success script"` in here?

greenboot-script-runner.sh
jlebon: I usually use `find` exactly because of the empty dir case. `ls` will error out.
E.g. something like: `find /etc/greenboot.d/required -name '*.sh'`

greenboot-script-runner.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+run_required_scripts () {
+  required_greenboot_scripts=`ls /etc/greenboot.d/required/*.sh`
+  set +e
+  for script in $required_greenboot_scripts; do
+    bash $script
```
jlebon: An alternative to `set +e` for that is:

```
rc=0
$script || rc=$?
```

Up to you which one to use!

greenboot.sh
```diff
@@ -1,9 +1,10 @@
...
-   echo "Greenboot Health Check: SUCCESS" | systemd-cat -p notice
+   echo "Health Check: SUCCESS"
 else
-   echo "Greenboot Health Check: FAILURE! Rolling back..." | systemd-cat -p emerg
-   # TODO: verify rpm-ostree status has more than one entry
-   rpm-ostree rollback
-   reboot
+   echo "Health Check: FAILURE! Starting rollback..." >&2
+   # TODO: verify rpm-ostree status has an older entry to boot back into
+   rpm-ostree rollback --reboot
```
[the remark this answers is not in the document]
dustymabe: i think we can leave here for now. and do redboot as an incremental goal

greenboot-script-runner.sh
LorbusChris: I like it

greenboot-script-runner.sh
LorbusChris: is "return code" (rc) ok?

LorbusChris: nvmd. exit code it is.

PR #5, opened 2018-06-14
Version 0.1

LorbusChris:


usr/lib/systemd/system/greenboot.service
```diff
@@ -0,0 +1,17 @@
...
+#  greenboot is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=greenboot Success Scripts Runner
+BindsTo=greenboot.target
+After=greenboot.target
+OnFailure=redboot.service
```
LorbusChris: @jlebon: This can be moved into the target. Where do you prefer it?

usr/lib/systemd/system/greenboot.service
```diff
@@ -0,0 +1,17 @@
+#  This file is part of greenboot.
+#
+#  greenboot is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=greenboot Success Scripts Runner
+BindsTo=greenboot.target
```
LorbusChris: This works with `Requires` too. Or `RequiredBy` on the target. BindsTo is a stronger binding. I was thinking it might make sense because once we have Watchdog check services and they make the target fail, this should be propagated to greenboot.service, too. BindsTo does that.

etc/greenboot.d/green/00_greenboot_notification.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "Health Check SUCCESS! Boot status is GREEN"
```
LorbusChris: Add command to update motd

etc/greenboot.d/red/00_redboot_notification.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "Health Check FAILURE! Boot status is RED"
```
LorbusChris: Add command to update motd

greenboot.spec
```diff
@@ -0,0 +1,103 @@
...
+mkdir -p %{buildroot}%{_sysconfdir}/%{name}.d/check/required
+install -Dpm 0755 etc/greenboot.d/check/required/00_required_scripts_start.sh %{buildroot}%{_sysconfdir}/%{name}.d/check/required/00_required_scripts_start.sh
+mkdir    %{buildroot}%{_sysconfdir}/%{name}.d/check/wanted
+install -Dpm 0755 etc/greenboot.d/check/wanted/00_wanted_scripts_start.sh %{buildroot}%{_sysconfdir}/%{name}.d/check/wanted/00_wanted_scripts_start.sh
+mkdir    %{buildroot}%{_sysconfdir}/%{name}.d/green
+install -Dpm 0755 etc/greenboot.d/green/00_greenboot_notification.sh %{buildroot}%{_sysconfdir}/%{name}.d/green/00_greenboot_notification.sh
+mkdir    %{buildroot}%{_sysconfdir}/%{name}.d/red
+install -Dpm 0755 etc/greenboot.d/red/00_redboot_notification.sh %{buildroot}%{_sysconfdir}/%{name}.d/red/00_redboot_notification.sh
+install -Dpm 0755 etc/greenboot.d/red/98_ostree_rollback.sh %{buildroot}%{_sysconfdir}/%{name}.d/red/98_ostree_rollback.sh
+install -Dpm 0755 etc/greenboot.d/red/99_reboot.sh %{buildroot}%{_sysconfdir}/%{name}.d/red/99_reboot.sh
```
LorbusChris: split this out into its own package, so we don't force everyone to reboot on a red boot.

usr/lib/systemd/system/greenboot-healthcheck.service
```diff
@@ -0,0 +1,17 @@
...
+[Unit]
+Description=greenboot Health Check Scripts Runner
+Before=greenboot.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/greenboot/greenboot.sh check
+
+[Install]
+RequiredBy=greenboot.target
```
jlebon: Hmm, this feels a bit "indirect". I would expect this dependency to happen directly in `greenboot.target` instead, e.g. something like:

```
Requires=greenboot-healthcheck.service
After=greenboot-healthcheck.service
```

?

usr/lib/systemd/system/greenboot.service
jlebon: This belongs in the target anyway, no? We should run `redboot.service` if we fail to run `check` scripts, not if we fail to run `green` scripts, right?

usr/libexec/greenboot/greenboot.sh
```diff
@@ -0,0 +1,76 @@
...
+  local rc=0
+  for script in $red_scripts; do
+    systemd-cat -t "$(basename $script)" bash $script || rc=$?
+    if [ $rc -eq 0 ]; then
+      echo -e "\e[1mRed Script '$(basename $script)' SUCCESS\e[0m"
+    else
+      echo -e "\e[1;31mRed Script '$(basename $script)' FAILURE (exit code '$rc'). Continuing...\e[0m" >&2
+    fi
+  done
+}
```
jlebon: Minor: we can probably dedup all this even more if we have a function that takes the path to scan, whether it's fatal, and whatever identifier we want to use when printing errors to the journal. But we can do that later!

etc/greenboot.d/check/wanted/00_wanted_scripts_start.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "Running greenboot Wanted Scripts"
+
+exit 0
```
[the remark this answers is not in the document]
LorbusChris: probably not

tests/Vagrantfile
```diff
@@ -0,0 +1,15 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "fedora/28-cloud-base"
```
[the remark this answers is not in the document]
LorbusChris: how to I enable copr repo with that?

usr/lib/systemd/system/greenboot.service
```diff
@@ -0,0 +1,17 @@
...
+After=greenboot.target
+OnFailure=redboot.service
+OnFailureJobMode=fail
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/greenboot/greenboot.sh green
```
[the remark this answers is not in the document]
LorbusChris: yes. have been doing it manually right now.

usr/lib/systemd/system/greenboot-healthcheck.service
LorbusChris: This would certainly work. Reasoning behind it was to keep the structure of this service just as any other, custom check service

usr/libexec/greenboot/greenboot.sh
LorbusChris: I had this in mind for when converting to C :) :+1:

usr/lib/systemd/system/greenboot.service
LorbusChris: ok, we'll use `Requires`, as discussed on IRC

usr/libexec/greenboot/greenboot.sh
LorbusChris: The color codes aren't picked up by systemd-cat :(

usr/lib/systemd/system/greenboot.target
```diff
@@ -0,0 +1,18 @@
+#  This file is part of greenboot.
+#
+#  greenboot is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=greenboot Success Target
+Wants=greenboot.service
```
LorbusChris: @jlebon re: https://github.com/LorbusChris/greenboot/pull/4#discussion_r195443658
change this to `Requires` to also fail the target on green script failures, not just health check failures, creating a circular dependency between both the service and the target.

greenboot.spec
LorbusChris: that's done

etc/greenboot.d/green/00_greenboot_notification.sh
LorbusChris: cc @dustymabe

tests/Vagrantfile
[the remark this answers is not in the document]
dustymabe: you don't but you can curl the repo file into /etc/yum.repos.d/. It's fine to leave it like it is for now.

etc/greenboot.d/green/00_greenboot_notification.sh
dustymabe: +1

PR #10, opened 2018-06-27
Add MotD updater, greenboot-motd package

### LINKED ISSUE #8: Notification scripts should update MOTD

LorbusChris:
Fixes #8

Please review @dustymabe @jlebon


README.md
```diff
@@ -6,19 +6,19 @@ The following directory structure is created:
...
-                      /required
-                      /wanted
-                /green
-                /red
+    /greenboot
+              /check
+                    /required.d
+                    /wanted.d
+              /green.d
+              /red.d
```
dustymabe: have you ever used `/usr/bin/tree` before? putting the lines in here like that utility does might be nice.

usr/libexec/greenboot/greenboot.sh
```diff
@@ -70,7 +70,7 @@ case "$@" in
     run_red_scripts
     ;;
   *)
-    echo -e "\e[31mIllegal Command\e[0m" >&2
```
dustymabe: were these color codes? Decided we didn't like them?

etc/greenboot.d/motd/redboot.motd
```diff
@@ -0,0 +1 @@
+Health Check FAILED! Boot Status is RED!
```
dustymabe: should we tell the user how to see which ones failed?

etc/greenboot.d/red/00_redboot_notification.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
-set -euo pipefail
-
-echo "Health Check FAILURE! Boot status is RED"
```
dustymabe: I wonder if this would still be useful (i.e. this would print in the journal right?)

README.md
LorbusChris: will check that out!

usr/libexec/greenboot/greenboot.sh
LorbusChris: They weren't picked up by the journal :( if there's a way to get them to work, I'd still like to have them very much

etc/greenboot/motd/redboot.motd
```diff
@@ -0,0 +1 @@
+Health Check FAILED! Boot Status is RED!
```
jlebon: Those aren't too useful yet though, given that the eventual goal is to reboot if we have a red boot. So you'd never actually see this, right?

Though I can picture it being useful if we develop this to show e.g. the number of reboots that were done before reaching a greenboot, or whether the default boot failed and we had to rollback, etc...

I guess what I think I'm saying is, we should eventually be generating `/run/greenboot/motd` dynamically? Instead of linking static files into there.

etc/greenboot.d/motd/redboot.motd
LorbusChris: good idea! Thinking about it, this message could also be generated dynamically to directly show what failed, similar to what Cockpit does for displaying its ip/url. will add a note for using the journal for now, and open an issue for a more dynamic generation of it.

etc/greenboot.d/red/00_redboot_notification.sh
LorbusChris: revive this as 00 then and make the motd update script 01?

etc/greenboot/motd/redboot.motd
LorbusChris: I guess so, sounds sensible

LorbusChris:
re openssh: https://bodhi.fedoraproject.org/updates/FEDORA-2018-73801e68f3

dustymabe:
haven't had a chance to review, but would love this in rpm form so I can play around with it tomorrow before our meeting. any chance I could get a build somewhere (or just merge this and we'll fixup any outstanding issues later)?

LorbusChris:
@dustymabe the dev branch builds live here: https://copr.fedorainfracloud.org/coprs/lorbus/greenboot-dev
I'll merge this now, though, so there'll be a build in the standard lorbus/greenboot copr in a minute too!

PR #14, opened 2018-07-13
Dedupe runner function

LorbusChris:
Add generic 'run' command

@jlebon @dustymabe

### DIFF
--- a/usr/libexec/greenboot/greenboot.sh
+++ b/usr/libexec/greenboot/greenboot.sh
@@ -2,75 +2,42 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-run_required_scripts () {
-  echo "Running Required Health Check Scripts..."
-  local required_scripts=`find /etc/greenboot/check/required.d -name '*.sh'`
-  local rc=0
-  for script in $required_scripts; do
+script_runner () {
+  local scripts_dir=$1; shift
+  local mode=$1; shift
+  local start_msg=$1; shift
+  echo "$start_msg"
+  for script in `find $scripts_dir -name '*.sh'`; do
+    local rc=0
     systemd-cat -t "$(basename $script)" bash $script || rc=$?
     if [ $rc -ne 0 ]; then
-      echo "Required Health Check Script '$(basename $script)' FAILURE (exit code '$rc')" >&2
-      exit $rc
-    fi
-    echo "Required Health Check Script '$(basename $script)' SUCCESS"
-  done
-}
-
-run_wanted_scripts () {
-  echo "Running Wanted Health Check Scripts..."
-  local wanted_scripts=`find /etc/greenboot/check/wanted.d -name '*.sh'`
-  local rc=0
-  for script in $wanted_scripts; do
-    systemd-cat -t "$(basename $script)" bash $script || rc=$?
-    if [ $rc -eq 0 ]; then
-      echo "Wanted Health Check Script '$(basename $script)' SUCCESS"
+      local failure_msg="Script '$(basename $script)' FAILURE (exit code '$rc')"
+      case "$mode" in
+        "relaxed")
+          echo "$failure_msg. Continuing..." >&2
+          ;;
+        "strict")
+          echo "$failure_msg" >&2
+          exit $rc
+      esac
     else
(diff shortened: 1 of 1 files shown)


usr/libexec/greenboot/greenboot.sh
```diff
@@ -2,63 +2,58 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-run_required_scripts () {
-  echo "Running Required Health Check Scripts..."
-  local required_scripts=`find /etc/greenboot/check/required.d -name '*.sh'`
+script_runner () {
+  if [ $# -eq 0 ]; then
+    echo "Please call '$0 run <dir> [{strict,relaxed}] [message]' to run this command!"
```
LorbusChris: s/dir/path is probably better

usr/libexec/greenboot/greenboot.sh
```diff
@@ -2,75 +2,56 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-run_required_scripts () {
-  echo "Running Required Health Check Scripts..."
-  local required_scripts=`find /etc/greenboot/check/required.d -name '*.sh'`
+script_runner () {
```
jlebon: One thing I usually do in bash functions is immediately assign the positional arguments to meaningful local vars. That makes it much easier to read. E.g.:

```bash
script_runner () {
  local scripts_dir=$1; shift
  local mode=$1; shift
  local message=$1; shift
  ...
}
```

usr/libexec/greenboot/greenboot.sh
```diff
@@ -2,75 +2,56 @@
...
+  if [ $# -eq 0 ]; then
```
jlebon: It's a bit odd to have this usage checking inside an implementation function. Let's maybe put it just before the main `case`?

usr/libexec/greenboot/greenboot.sh
```diff
@@ -2,75 +2,56 @@
...
+    echo "Please call '$0 run <dir_path> [{strict,relaxed}] [message]' to run this command!"
+    exit 1
+  fi
   local rc=0
-  for script in $required_scripts; do
+  local scripts=`find $1 -name '*.sh'`
+  case "${2:-s}" in
```
jlebon: Honestly, it's probably better to just let callers spell it out the long way (`strict` or `relaxed`) to make it more self-documenting.

usr/libexec/greenboot/greenboot.sh
```diff
@@ -2,75 +2,56 @@
...
+    "r" | "relaxed")
+      local mode=relaxed
+      ;;
+    "s" | "strict" | *)
+      local mode=strict
+      ;;
+  esac
+  local start_msg=${3:-Running scripts from $1 in $mode mode...}
+  echo "$start_msg"
+  for script in $scripts; do
```
jlebon: You'll want to make sure you always reset `rc` here before running the script. Otherwise, the first fail will mark all following scripts as fails too!

LorbusChris: whoops!

usr/libexec/greenboot/greenboot.sh
LorbusChris: ack

usr/libexec/greenboot/greenboot.sh
LorbusChris: I'll put this into the "run" case

LorbusChris:
@jlebon: I adressed your review.

usr/libexec/greenboot/greenboot.sh
```diff
@@ -2,75 +2,57 @@
...
     ;;
   "green")
-    run_green_scripts
+    script_runner "/etc/greenboot/green.d" "relaxed" "Running Green Scripts..."
     ;;
   "red")
-    run_red_scripts
+    script_runner "/etc/greenboot/red.d" "relaxed" "Running Red Scripts..."
+    ;;
+  "run")
```
jlebon: Do we need to support a generic `run`? It feels like the other dirs should cover all use cases, right?

dustymabe:
If jlebon is +1 I'm :+1:

PR #15, opened 2018-07-23
Add priority prefixes to echo messages

LorbusChris:
Adds boldness/color to their corresponding journal logs.

What priority level do you feel is adequate for which case? @jlebon @dustymabe @nullr0ute 

from syslog(3), levels 0-7:
```
       This determines the importance of the message.  The levels are, in
       order of decreasing importance:

       LOG_EMERG      system is unusable

       LOG_ALERT      action must be taken immediately

       LOG_CRIT       critical conditions

       LOG_ERR        error conditions

       LOG_WARNING    warning conditions

       LOG_NOTICE     normal, but significant, condition

       LOG_INFO       informational message

       LOG_DEBUG      debug-level message
```

### DIFF
--- a/etc/greenboot/green.d/00_greenboot_notification.sh
+++ b/etc/greenboot/green.d/00_greenboot_notification.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -euo pipefail
 
-echo "Health Check SUCCESS! Boot status is GREEN"
+echo "<5>Health Check SUCCESS! Boot status is GREEN"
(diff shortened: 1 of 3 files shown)


etc/greenboot/green.d/00_greenboot_notification.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -euo pipefail
 
-echo "Health Check SUCCESS! Boot status is GREEN"
+echo "<5>Health Check SUCCESS! Boot status is GREEN"
```
dustymabe: does prepending with `<level>` really make it print out a message at that level ?

jlebon:
No major qualms with the levels chosen. Given that we might reboot as a result of a failed script, `LOG_EMERG` makes sense I think. That also means a higher chance of being visible on tty0/wall messages.

PR #16, opened 2018-07-25
[WIP] Add boot counting

LorbusChris:
RFC @jlebon @dustymabe @nullr0ute
Depends on https://github.com/rhboot/grub2/pull/24


etc/greenboot/green.d/00_grub_fallback.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+# We need to determine whether this boot was a fallback boot
+if grub2-editenv list | grep -q "fallback_active=true"; then
+  grub2-editenv - unset fallback_active
```
LorbusChris: we could as well leave this set until next rpm-ostree upgrade

etc/greenboot/green.d/00_grub_fallback.sh
```diff
@@ -0,0 +1,12 @@
...
+  grub2-editenv - set boot_success=0 # this should already be 0
+  grub2-editenv - set boot_counter=2
+  rpm-ostree rollback --reboot
```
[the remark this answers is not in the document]
dustymabe: @jlebon, do we need to run `rollback` here? or just some form of `rpm-ostree cleanup` ?

etc/greenboot/green.d/00_grub_fallback.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+# We need to determine whether this boot was a fallback boot
+if grub2-editenv list | grep -q "fallback_active=true"; then
```
jlebon: Looks like this should be `boot_counter=-1` now?
Also let's wrap regex with `^` and `$` to be safe.

etc/greenboot/green.d/00_grub_fallback.sh
jlebon: I think leaving it set makes more sense yeah. We can use that information elsewhere as well to let users know this is a fallback boot.

etc/greenboot/green.d/00_grub_fallback.sh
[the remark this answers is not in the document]
jlebon: It depends on the behaviour we want I suppose. I'm slightly in favour of rollback to keep the bad deployment around in case admins want to prod it manually. A later upgrade will wipe it out. Though `cleanup` is fine too, assuming we at least write down information (e.g. version, checksum, timestamp) about the deployment that failed somewhere before removing it? Anyway, we'll likely need to revisit this once time comes to integrate this with rpm-ostree, where we'll want to make sure we don't try to auto-update to the same broken deployment again.

Either way, for now it seems like we want something like this?

```
rpm-ostree rollback
grub2-editenv - set boot_success=1
```

Then we don't even need to reset `boot_success` and `boot_counter`, right?

etc/greenboot/green.d/00_grub_fallback.sh
```diff
@@ -0,0 +1,12 @@
...
+  grub2-editenv - unset fallback_active
+  grub2-editenv - set boot_success=0 # this should already be 0
+  grub2-editenv - set boot_counter=2
```
jlebon: If we're doing a rollback, then we want to set `boot_success=1` so that the first entry (using 1-base index here :)) is always picked right? (Where the first entry is now the entry we're booted into when we're executing this).

etc/greenboot/green.d/00_grub_fallback.sh
```diff
@@ -0,0 +1,12 @@
...
+  rpm-ostree rollback --reboot
+else
+  grub2-editenv - set boot_success=1
```
jlebon: So in fact, it seems like we're doing this either way -- so we can factor that out of the if statement.

PR #18, opened 2018-08-10
Dynamic MotD

LorbusChris:


LorbusChris:
@jlebon @dustymabe if one of you could have a look, that'd be great!

etc/greenboot/green.d/50_greenboot_motd.sh
```diff
@@ -1,4 +1,16 @@
 #!/bin/bash
 set -euo pipefail
 
-ln -snf /etc/greenboot/motd/greenboot.motd /run/greenboot/motd
+if [ ! -d /etc/motd.d ]; then
+    mkdir /etc/motd.d
+fi
+
+greenboot_logs=`journalctl -u greenboot.service -p 0..5 -b -0 -o cat`
```
dustymabe: nit: I think generally `$(command arg)` is preferred over `'command arg'` (using backticks)

etc/greenboot/motd/redboot.motd
```diff
@@ -1,2 +0,0 @@
-Health Check FAILED! Boot Status is RED!
-Run `journalctl -u greenboot-healthcheck` for more information.
```
dustymabe: why not update motd with redboot status?

etc/greenboot/green.d/50_greenboot_motd.sh
```diff
@@ -1,4 +1,16 @@
...
+printf '%s\n' "$greenboot_logs" > /etc/greenboot/motd
```
dustymabe: are we writing to `/etc/greenboot/motd` on every boot? seems like we should be writing to somehwere in /run/ instead?

dustymabe:
cc @rfairley since you just learned a lot of about motd and are [working on enhancing motd.d functionality](https://github.com/linux-pam/linux-pam/issues/68)

etc/greenboot/motd/redboot.motd
LorbusChris: the reason I changed it was that (with auto rollback enabled) you'd never actually see it. Thinking about it a bit more, it probably makes sense to have it, for people not running ostree systems.
Do you think it thus makes sense to split the motd-updater out into its own service unit which is wanted by both greenboot.service and redboot.service units?

etc/greenboot/green.d/50_greenboot_motd.sh
LorbusChris: with @rfairley's changes, we could probably do this without symlinking and instead writing directly to `/run/motd.d/greenboot`

rfairley: yes - hopefully when the changes land the symlink can be avoided - hope to get something up for review soon.

I do something similar in the `motd` [package I had been working on](https://github.com/rfairley/fedora-coreos-login-messages) - [writes a new motd](https://github.com/rfairley/fedora-coreos-login-messages/blob/master/usr/lib/coreos/motdgen#L66) to `/run/${%name}.motd` with it symlinked from `/etc/motd.d`. With the PAM changes `/run/motd.d` should become a "canonical" directory to do this for any package.

jlebon:
Looks sane from a quick scan! I'll defer to @dustymabe and @rfairley. :)

etc/greenboot/green.d/50_greenboot_motd.sh
rfairley: I guess as of now, writing to `/etc` is the best option,  as symlinks to `/run` alone won't work due to the modifications needed in the selinux policy for PAM to read the file in `/run`. But the changes in PAM (which will also bring the modified selinux policy for PAM ot read `/run/motd.d`) should avoid the need in the future to make a symlink to run.

I think the writing to `/etc` is good to merge for now - to not block on the PAM changes

etc/greenboot/motd/redboot.motd
dustymabe: > Thinking about it a bit more, it probably makes sense to have it, for people not running ostree systems

yeah or people who somehow have an ostree system but don't want "automatic" rollback, but want to investigate first for whatever reason.

> Do you think it thus makes sense to split the motd-updater out into its own service unit which is wanted by both greenboot.service and redboot.service units?

I'm not sure I have an opinion on that. Sounds reasonable to me.

etc/greenboot/green.d/50_greenboot_motd.sh
dustymabe: +1

dustymabe:
@LorbusChris - are you still waiting on us for review here or are we waiting on you for changes?

LorbusChris:
@dustymabe you're waiting on me :) I didn't prioritize this, will come around to it tomorrow I presume

LorbusChris:
@dustymabe done.
Should I go ahead an merge or wait for https://github.com/linux-pam/linux-pam/pull/69 to land first, which'll make this doable without any symlink?

dustymabe:
> Should I go ahead and merge or wait for [linux-pam/linux-pam#69](https://github.com/linux-pam/linux-pam/pull/69) to land first, which'll make this doable without any symlink?

I wouldn't block on that, since we're not sure when it will get merged. Maybe make a note to check it periodically so we can change it once it does get merged :)

Issue #20, opened 2018-08-14
Track systemd's boot-complete.target feature


LorbusChris:
https://github.com/systemd/systemd/pull/9437

LorbusChris:
boot-complete.target has been added in upstream systemd, time to have a closer look at it!

PR #21, opened 2019-02-05
Update to v0.6

LorbusChris:
These are more changes than originally anticipated for the 0.6 Release.
I however decided to go ahead and reshape this to completely get rid of greenboot.target.
The scripts have been renamed to match the service names, which makes for a better experience when looking at the journal logs.

Changelog:

- Remove `greenboot.target` and instead integrate with systemd's `boot-complete.target`
- Ship default fallback MotD file in `/usr/lib/motd.d/boot-status`
- ~~Write generated MotD to `/run/motd.d/boot-status` instead of symlinking to `/etc/motd.d/bootstatus`~~
- Renamed `greenboot.sh` script to `greenboot`
- Renamed `greenboot_motdgen.sh` script to `greenboot-status`
- Renamed `greenboot-motdgen.service` unit file to `greenboot-status.service`

Resolves: #20 
~~Depends on: https://github.com/fedora-selinux/selinux-policy/pull/244~~ (Moved to 0.7)


tests/run_test.sh
```diff
@@ -0,0 +1,30 @@
...
+# sudo restorecon -Rv /etc/libvirt # -> Relabeled /etc/libvirt from system_u:object_r:virt_etc_rw_t:s0 to system_u:object_r:virt_etc_t:s0
+
+cd atomic
+
+vagrant up
+# Use the failing check to test red boot status behaviour
+# vagrant ssh -c 'ln -s /home/vagrant/sync/10_failing_check.sh /etc/greenboot/check/required.d/10_failing_check.sh'
+
+# Workaround for https://github.com/fedora-selinux/selinux-policy/issues/242
+vagrant ssh -c 'sudo semanage fcontext -a -t pam_var_run_t "/var/run/motd.d(/.*)?"; \
```
rfairley: Minor thing, but this regex should be updated to the corrected form `/var/run/motd\.d(/.*)?` (https://github.com/fedora-selinux/selinux-policy/pull/244#discussion_r250683367).

greenboot.spec
```diff
@@ -21,7 +23,7 @@ Requires:           systemd
 %package motd
 Summary:            Message of the Day updater for greenboot
 Requires:           %{name} = %{version}-%{release}
-Requires:           pam >= 1.3.1
+Requires:           pam >= 1.3.1-15
```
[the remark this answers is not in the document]
rfairley: +1 on adding the `-15` and adding a selinux-policy Requires once https://github.com/fedora-selinux/selinux-policy/pull/244 is in.

I think there will also be a Requires on `setup >= 2.12.7` once this gets merged https://pagure.io/setup/pull-request/16# - to make sure the `/run/motd` and `/run/motd.d` paths are already created so that the proper context can be applied.

usr/libexec/greenboot/greenboot-status
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+if [ ! -f /run/motd.d/boot-status ]; then
+    if [ ! -d /run/motd.d ]; then
+        mkdir /run/motd.d
+    fi
+    touch /run/motd.d/boot-status
```
rfairley: On creating files in `/run/motd.d/*`, it occurred to me that since any package can write here, packages should own the file that they write there in the specfile (with a tmpfile to create the `/run/motd.d/mypkgmotd` file). I'm planning to do this with the `/run/motd.d/console-login-helper-messages.motd` that gets generated ([rough issue for this](https://github.com/rfairley/console-login-helper-messages/issues/24)).

Doing the check `if [ ! -f /run/motd.d/boot-status ]; then` already makes this safe though, so I don't think the ownership of `/run/motd.d/boot-status` should block this PR. Just something to keep in mind.

LorbusChris:
@nullr0ute currently pulling a Vagrant box to test, seems to take a while.

Any obvious things I missed here in the meantime?

PR #22, opened 2019-02-13
v0.7

LorbusChris:
Ready to test


LorbusChris:
@dustymabe @jlebon @nullr0ute @cgwalters I know this change is big, but if you could have a look, that'd be great!

usr/libexec/greenboot/greenboot
```diff
@@ -21,7 +21,7 @@ script_runner () {
           exit $rc
       esac
     else
-      echo "<6>Script '$(basename $script)' SUCCESS"
+        echo "<6>Script '$(basename $script)' SUCCESS"
```
LorbusChris: oops

usr/lib/systemd/system/greenboot-healthcheck.service
```diff
@@ -7,10 +9,13 @@
 
 [Unit]
 Description=greenboot Health Checks Runner
+After=multi-user.target
 Before=boot-complete.target
+SyslogIdentifier=greenboot-healthcheck
 
 [Service]
 Type=oneshot
+RemainAfterExit=yes
```
LorbusChris: is this needed?

tests/run_test.sh
```diff
@@ -34,7 +22,9 @@ vagrant ssh -c 'sudo rpm-ostree cleanup -m; \
                 sudo grub2-editenv - set boot_counter=1; \
                 echo "rebooting..." && sudo systemctl reboot;'
 
-vagrant ssh -c 'sudo journalctl -u greenboot-healthcheck -u boot-complete.target -u greenboot -u redboot -u greenboot-status'
+# vagrant ssh -c 'sudo systemctl start rpm-ostreed-automatic.service'
```
LorbusChris: remove line

usr/lib/systemd/system/greenboot-grub2-set-counter.service
```diff
@@ -0,0 +1,20 @@
...
+#  (at your option) any later version.
+
+[Unit]
+Description=Set grub2 boot counter in preparation of upgrade
+Before=ostree-finalize-staged.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/greenboot/greenboot-grub2-set-counter 3
```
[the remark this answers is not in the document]
jlebon: Right, so another thing here is the bit I mentioned in https://github.com/projectatomic/rpm-ostree/issues/1758#issuecomment-463797394:

> (though you'd still want to check that a new deployment was created)

Basically, you don't want to always do this on each reboot/shutdown; only on those where `ostree-finalize-staged.service` actually did something. Which argues for adding this hook functionality directly in OSTree proper. For the time being though, what you could do is have an `ExecStart` that registers the current default deployment somewhere in `/run`, and then the `ExecStop` script only sets the counter if the default deployment changed.

greenboot.spec
```diff
@@ -22,8 +22,10 @@ Requires:           systemd
 %package status
 Summary:            Message of the Day updater for greenboot
 Requires:           %{name} = %{version}-%{release}
+Requires:           setup >= 2.12.7
```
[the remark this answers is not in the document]
LorbusChris: the motd selinux labels need at least that build

usr/lib/systemd/system/greenboot-grub2-set-counter.service
```diff
@@ -0,0 +1,21 @@
...
+#
+#  greenboot is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=Set grub2 boot counter in preparation of upgrade
+Before=ostree-finalize-staged.service
+ConditionPathExists=/run/ostree/staged-deployment
```
LorbusChris: @jlebon is this file here when we install or override arbitrary packages? I think this way we can make sure we only set the grub countdown when we actually need to.

jlebon: Yeah, it'll exist anytime there's a new deployment we created. To contrast, note that e.g. `rpm-ostree rollback` doesn't create a deployment but just re-arranges existing ones.

jlebon: (Just had a look at the service file and it looks good overall!)

PR #28, opened 2020-04-01
Update to v0.9

LorbusChris:
- greenboot-status: Improve status output
- greenboot: Sort scripts before running with script_runner
- greenboot: Split into service and target
- redboot-auto-reboot: Guard against boot looping even more
- greenboot-grub2-set-counter: Break dependency cycle

To stop grub-boot-success.timer from interfering with our setting of the boot_success grubenv var, we also need:
- https://src.fedoraproject.org/rpms/fedora-release/pull-request/117 (f33)
- https://src.fedoraproject.org/rpms/fedora-release/pull-request/118 (f32)


LorbusChris:
@nullr0ute this is good to go. I've tested the behavior on a VM, and everything seems to work well now.

The PRs to add user presets are also linked. They would disable grub''s grub_boot_success.timer so it stops interfering with greenboot setting that. That timer is however only activated when a user logs in, so if nobody logs in it won't set it and won't interfere.

LorbusChris:
For RHEL, it should be ensured that all presets for IoT from fedora-release are shipped somehow.

nullr0ute:
@LorbusChris OK, can we get it tagged and a build in Fedora? We should also look at packit auto build bits, I can do a PR for that for the next release.

LorbusChris:
I can do a new koji build once this merges.

Is Packit ready yet to trigger Koji builds? Or is that just COPR builds as CI tests?

PR #31, opened 2020-05-26
Fix onFailure not being respected on greenboot.target

### LINKED ISSUE #30: systemd unit file warning: OnFailure=redboot.target ignored
`greenboot.target: Requested dependency OnFailure=redboot.target ignored (target units cannot fail).`

This is due to a [change](https://github.com/systemd/systemd/commit/c80a9a33d04fb4381327a69ce929c94a9f1d0e6c) in systemd, included in v245.

I am not sure if that actually breaks the functionality but it would appear so, since it should trigger the red path for boot failures.

LorbusChris:
Fixes #30


LorbusChris:
xref: https://src.fedoraproject.org/rpms/fedora-release/pull-request/128

PR #33, opened 2020-06-11
Add packit support

nullr0ute:
Sync the spec file from Fedora, and add the packit config so we can get
automated PRs, CI, and associated bits for packaging.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,18 @@
+specfile_path: greenboot.spec
+synced_files:
+  - greenboot.spec
+  - .packit.yaml
+upstream_package_name: greenboot
+downstream_package_name: greenboot
+jobs:
+  - job: sync_from_downstream
+    trigger: commit
+  - job: propose_downstream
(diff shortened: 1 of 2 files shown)


LorbusChris:
LGTM

LorbusChris:
We'll need to enable https://github.com/marketplace/packit-as-a-service on this repo/org.

I believe recently there was support for building right in Fedora's Koji added to packit as well. AFAIK it will require the `packit` Fedora user to be added to the greenboot RPM maintainers for it to be able to do merges and trigger new builds.

nullr0ute:
The way I've seen it work is that without the access it files PRs once it's done test builds etc

LorbusChris:
I think even for that to work (i.e. PR builds on COPR, and for submitting PRs for upstream/downstream syncing) we need to enable the packit service on this repo. (This may have changed in the meantime and not be a correct assumption anymore, though.)

PR #34, opened 2020-08-12
v0.11.0: Fix race between checking for fallback and setting success

LorbusChris:
https://github.com/osbuild/osbuild-composer/pull/893#issuecomment-673102281

### DIFF
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -1,6 +1,6 @@
 Name:               greenboot
-Version:            0.10.3
-Release:            2%{?dist}
+Version:            0.11.0
+Release:            1%{?dist}
 Summary:            Generic Health Check Framework for systemd
 License:            LGPLv2+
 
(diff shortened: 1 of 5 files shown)


nullr0ute:
LGTM!

LorbusChris:
Let's old off on this. I'll change the set success service to not require greenboot-task-runner enabled, but instead systemd's bless-boot. Will require preset changes. Will look into it at the start of next week.

LorbusChris:
Ok, upon a little bit of investigation it doesn't seem to me like the bless boot service can be reused as that is geared towards usage with systemd-boot.
Instead, I've changed the set-success to be required by multi-user.target so it is started even if greenboot-task-runner is disabled. It still requires boot-complete.target first, so the timing should still be correct.

I've also removed the ants=redboot-task-runner from the auto-reboot service so the reboot-task-runner is not started by the auto-reboot service if it's disabled. If it is enabled, the redboot-task-runner will still be started before, by redboot.target. 

This should be good to go now.

PR #36, opened 2021-06-11
Requirement of systemd version >= 240 set in spec file

### LINKED ISSUE #35: spec: requires at least system 240
The requirement for `systemd` is currently too vague, since greenboot requires `boot-complete.target`, [added](https://github.com/systemd/systemd/commit/329d20db3cb02d789473b8f7e4a59526fcbf5728) with systemd 240.

jnogol:
As @gicmo explained in #35, Greenboot requires `boot-complete.target`, which was included in [systemd 240](https://github.com/systemd/systemd/commit/329d20db3cb02d789473b8f7e4a59526fcbf5728). Hence, specifying this requirement in the spec file.

Fixes #35.

### DIFF
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -14,7 +14,7 @@ Source0:            https://github.com/%{repo_owner}/%{repo_name}/archive/%{repo
 BuildArch:          noarch
 BuildRequires:      systemd-rpm-macros
 %{?systemd_requires}
-Requires:           systemd
+Requires:           systemd >= 240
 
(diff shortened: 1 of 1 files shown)


LorbusChris:
LGTM, but please make sure this change is also made in https://src.fedoraproject.org/rpms/greenboot/blob/rawhide/f/greenboot.spec which is the canonical location of the specfile.

nullr0ute:
LGTM!

PR #42, opened 2021-06-28
#41 Maximum number of boot attempts configurable

### LINKED ISSUE #41: [Feature] - Add ability to configure maximum number of boot attempts
Following up on https://bugzilla.redhat.com/show_bug.cgi?id=1920063,

Although it can be set via https://github.com/fedora-iot/greenboot/blob/main/usr/libexec/greenboot/greenboot-grub2-set-counter, it is hardcoded in https://github.com/fedora-iot/greenboot/blob/main/usr/lib/systemd/system/greenboot-grub2-set-counter.service#L19. Extending the former so it can be configured via environment variable as well can be interesting.

jnogol:
Add ability to configure maximum number of boot attempts via env var and config file.

Closes #41

### DIFF
--- a/usr/libexec/greenboot/greenboot-grub2-set-counter
+++ b/usr/libexec/greenboot/greenboot-grub2-set-counter
@@ -1,8 +1,19 @@
 #!/bin/bash
-set -euo pipefail
+set -eo pipefail
 
-max_boot_attempts=${1:-3} # default to 3 attempts
+GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/greenboot.conf
+if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
+    source $GREENBOOT_CONFIGURATION_FILE    
+fi
+
+if [ -n "$1" ]; then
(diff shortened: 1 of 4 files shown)


usr/libexec/greenboot/greenboot-grub2-set-counter
```diff
@@ -1,8 +1,19 @@
 #!/bin/bash
-set -euo pipefail
+set -eo pipefail
 
-max_boot_attempts=${1:-3} # default to 3 attempts
+GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/config
```
LorbusChris, line 4:
```suggestion
GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/greenboot.conf
```

LorbusChris (changes requested):
Let's make the name a bit more descriptive. Also, since we're effectively introducing that new config file here, let's also document it

LorbusChris:
LGTM

PR #50, opened 2021-09-07
Add watchdog-triggered boot check.

jnogol:
Addresses #2

`/etc/greenboot/check/required.d/02_watchdog.sh` checks whether the current boot was watchdog-triggered or not and, if it has occured within a certain user-defined grace period (in hours), it'll mark the boot as `red`.

### DIFF
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -45,6 +45,15 @@ Obsoletes:          greenboot-update-platforms-check <= 0.12.0
 %description default-health-checks
 %{summary}.
 
+%package watchdog-triggered-boot-check
+Summary:            Watchdog-triggered boot check for greenboot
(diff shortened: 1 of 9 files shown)


greenboot.spec
```diff
@@ -36,8 +36,10 @@ Obsoletes:          greenboot-rpm-ostree-grub2 <= 0.12.0
 %{summary}.
 
 %package default-health-checks
-Summary:            Update platforms DNS resolution and connection check for greenboot
+Summary:            Series of optional and curated health checks
```
jnogol, line 39: @LorbusChris @nullr0ute what do you think about this summary?

jnogol:
I think Packit has cached some stuff regarding the specfile again, will open a new PR to see if we can solve this problem.

PR #54, opened 2021-09-16
Move default health checks to read-only location

### LINKED ISSUE #51: Move default health checks to a read-only directory
In order to avoid accidental modifications of default health checks, we should move those to a read-only directory.

This implies that Greenboot should be able to check two directories now, the read-only one for the default scripts and the current one (`/etc/greenboot/check/{required,wanted}.d`).

jnogol:
Closes #51


Dockerfile
```diff
@@ -1,12 +1,12 @@
...
-RUN dnf install -y git findutils systemd grub2-tools-minimal
+RUN dnf install -y git findutils systemd grub2-tools-minimal util-linux
 
 RUN git clone https://github.com/bats-core/bats-core.git
 WORKDIR /bats-core
 RUN ./install.sh /usr/local
 
 COPY ./usr/libexec/greenboot /usr/libexec/greenboot
-COPY ./etc/greenboot/check /etc/greenboot/check
+COPY ./usr/etc/greenboot/check /usr/etc/greenboot/check
```
LorbusChris, line 9: I think this should be moved to `/usr/lib/`, not `/usr/etc/`

jnogol: You're right. Updated.

jnogol:
Need to add tests that cover the two directories feature. Plus, will remove the relationship of this PR with #50, so they can be reviewed independently.

greenboot.spec
```diff
@@ -170,14 +174,14 @@ install -DpZm 0755 etc/greenboot/check/wanted.d/* %{buildroot}%{_sysconfdir}/%{n
...
-%dir %{_sysconfdir}/%{name}/green.d
-%dir %{_sysconfdir}/%{name}/red.d
+%dir %{_prefix}/lib/%{name}
+%dir %{_prefix}/lib/%{name}/check
+%dir %{_prefix}/lib/%{name}/check/required.d
+%{_prefix}/lib/%{name}/check/required.d/00_required_scripts_start.sh
+%dir %{_prefix}/lib/%{name}/check/wanted.d
+%{_prefix}/lib/%{name}/check/wanted.d/00_wanted_scripts_start.sh
+%dir %{_prefix}/lib/%{name}/green.d
+%dir %{_prefix}/lib/%{name}/red.d
```
LorbusChris, line 184: Let's leave the (noe empty) dirs in `/etc` as part of the package, so they get created and can be traced back to the RPM

```suggestion
%dir %{_prefix}/lib/%{name}
%dir %{_prefix}/lib/%{name}/check
%dir %{_prefix}/lib/%{name}/check/required.d
%{_prefix}/lib/%{name}/check/required.d/00_required_scripts_start.sh
%dir %{_prefix}/lib/%{name}/check/wanted.d
%{_prefix}/lib/%{name}/check/wanted.d/00_wanted_scripts_start.sh
%dir %{_prefix}/lib/%{name}/green.d
%dir %{_prefix}/lib/%{name}/red.d
%dir %{_sysconfdir}/%{name}
%dir %{_sysconfdir}/%{name}/check
%dir %{_sysconfdir}/%{name}/check/required.d
%dir %{_sysconfdir}/%{name}/check/wanted.d
%dir %{_sysconfdir}/%{name}/green.d
%dir %{_sysconfdir}/%{name}/red.d
```

LorbusChris (changes requested):
Also: Is the change in access permissions of ` usr/libexec/greenboot/greenboot-rpm-ostree-grub2-check-fallback` from 644 to 755 intended?

greenboot.spec
jnogol: > Also: Is the change in access permissions of usr/libexec/greenboot/greenboot-rpm-ostree-grub2-check-fallback from 644 to 755 intended?

It was the only file in that directory that had 644, not that we've had an issue or bug with it, just for a matter of consistency.

Regarding the suggestion, sounds fantastic, will commit it.

jnogol:
@LorbusChris It was the only file in that directory that had 644, not that we've had an issue or bug with it, just for a matter of consistency.

jnogol:
Regarding the fails in the packit builds, it seems the watchdog package was cached and as I have removed that commit from this PR, it's making the build fail.

jnogol:
Closing and opening a new PR to avoid Packit cache problems.

PR #55, opened 2021-09-20
 Move default health checks to read-only location

### LINKED ISSUE #51: Move default health checks to a read-only directory

jnogol:
Closes #51


LorbusChris (approved):
LGTM

nullr0ute (approved):
LGTM

PR #56, opened 2021-09-21
Ensure all required health checks are run

### LINKED ISSUE #52: Greenboot stops checking required health checks if one fails
As $subject states, this behaviour makes sense as Greenboot would mark the boot as "red/faulty" anyway, but it would bring more value to the user knowing all the required health checks that had failed, not just the first one.

jnogol:
Closes #52

### DIFF
--- a/usr/libexec/greenboot/greenboot
+++ b/usr/libexec/greenboot/greenboot
@@ -9,6 +9,7 @@ script_runner () {
   local scripts_dir=$1; shift
   local mode=$1; shift
   local start_msg=$1; shift
+  local required_hc_failed=false
   echo "$start_msg"
   for script in $(find "$scripts_dir" -name '*.sh' | sort); do
     local rc=0
@@ -21,13 +22,17 @@ script_runner () {
           echo "<2>$failure_msg. Continuing..." >&2
           ;;
         "strict")
-          echo "<0>$failure_msg" >&2
-          exit $rc
+          required_hc_failed=true
+          echo "<0>$failure_msg. Continuing..." >&2
       esac
     else
       echo "<6>Script '$(basename "$script")' SUCCESS"
     fi
   done
+
+  if [[ $required_hc_failed == true ]]; then
+    exit 1
+  fi
 }
 
 case "$1" in
--- a/tests/greenboot_check_fail_required.bats
+++ b/tests/greenboot_check_fail_required.bats
@@ -1,7 +1,8 @@
 load common.bash
 
 function setup() {
-    cp testing_files/10_failing_check.sh $GREENBOOT_DEFAULT_CHECK_PATH/required.d/
+    cp testing_files/10_failing_check.sh $GREENBOOT_DEFAULT_CHECK_PATH/required.d/10_failing_check.sh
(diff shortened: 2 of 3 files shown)


usr/libexec/greenboot/greenboot
```diff
@@ -7,6 +7,7 @@ script_runner () {
   local scripts_dir=$1; shift
   local mode=$1; shift
   local start_msg=$1; shift
+  local required_hc_failed="false"
```
LorbusChris, line 10: I don't think quotation marks are necessary here

jnogol: You're right. Pushing changes...

usr/libexec/greenboot/greenboot
```diff
@@ -19,13 +20,17 @@ script_runner () {
...
+          required_hc_failed=true
+          echo "<0>$failure_msg. Continuing..." >&2
       esac
     else
       echo "<6>Script '$(basename "$script")' SUCCESS"
     fi
   done
+
+  if [[ $required_hc_failed == true ]]; then
+    exit 1
```
LorbusChris, line 32: maybe echo something here, too, noting some (at least one) tests have failed?

jnogol: Good idea.

jnogol: What do you think @LorbusChris?

journalctl:
```
Oct 05 08:26:14 fedora 00_required_scripts_start.sh[681]: Running greenboot Required Health Check Scripts
Oct 05 08:26:14 fedora greenboot[670]: Script '00_required_scripts_start.sh' SUCCESS
Oct 05 08:26:14 fedora greenboot[670]: Script '01_fail.sh' FAILURE (exit code '1'). Continuing...
Oct 05 08:26:15 fedora greenboot[670]: Script '02_fail.sh' FAILURE (exit code '1'). Continuing...
Oct 05 08:26:15 fedora zsh.sh[694]: ZSH is available
Oct 05 08:26:15 fedora greenboot[670]: Script 'zsh.sh' SUCCESS
Oct 05 08:26:15 fedora greenboot[670]: ATTENTION: At least one required health check has failed
```

MOTD:
```
Script '01_fail.sh' FAILURE (exit code '1'). Continuing...
Script '02_fail.sh' FAILURE (exit code '1'). Continuing...
ATTENTION: At least one required health check has failed
Boot Status is RED - Health Check FAILURE!
```

If OK, will commit and update the PR.

LorbusChris: LGTM, although attention would sound like a warning to me, not an error. Not too important here though I think.

jnogol: Yeah, I wasn't too sure about that choice of words either but couldn't think of a better one. @nullr0ute, any feedback?

nullr0ute: We should possibly also report which of the health checks failed so the user knows where to look.

LorbusChris: That should already be the case here (e.g. `Script '02_fail.sh' FAILURE (exit code '1'). Continuing...` in the MOTD above)

PR #58, opened 2021-10-26
WIP: Systemd services health check

jnogol:
WIP PR: will squash the commits once everything's done.

**Goal**
A user will expect certain services such as NetworkManager and sshd to work in all circumstances and if they fail on an upgrade we should automatically rollback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,5 +11,6 @@ COPY ./etc/greenboot/check /etc/greenboot/check
 WORKDIR /testing
 COPY ./tests .
 COPY ./tests/testing_files/fedora_iot.conf /etc/ostree/remotes.d/fedora_iot.conf
+COPY ./tests/testing_files/greenboot.conf /etc/greenboot/greenboot.conf
 
 ENTRYPOINT [ "/bin/bash", "launch_all_tests.sh" ]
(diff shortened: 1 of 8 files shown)


jnogol:
I've tried this but for some reason it hadn't worked:

failing.service (similar to [this one](https://github.com/fedora-iot/greenboot/blob/main/development/failing.service))
```
[Unit]
Description=greenboot Failing Check Example
OnFailure=greenboot-check-if-boot-complete.service

[Service]
Type=oneshot
ExecStart=/bin/false

[Install]
WantedBy=multi-user.target
```

then, greenboot-check-if-boot-complete.service

```
[Unit]
Description=greenboot check if boot complete
OnFailure=redboot-auto-reboot.service
OnFailureJobMode=fail

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/etc/greenboot/greenboot-check-if-boot-complete

[Install]
WantedBy=multi-user.target
```

And it's not working. I think that the problem is that `redboot-auto-reboot-service` runs `Before=redboot.target` and this target `conflicts=boot-complete.target` which is already achieve.

What do you think? @LorbusChris

LorbusChris:
I wonder if we could achieve something like this by comparing the output of `systemctl list-units --failed` against a list of required services, and fail the boot in case of a match.

LorbusChris:
Thanks for your work on this @jnogol. I'm going to close this PR now as it's been superseded by #70

PR #59, opened 2021-11-02
Minor spec fixes

nullr0ute:
Fix the version of pam we actually require, fix up the provides.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


jnogol:
LGTM

PR #64, opened 2021-11-17
README updated with TOC and improved explanations

jnogol:
Summary:
- Reorganised sections so more important ones for the user are in the top.
- Added a TOC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,47 +1,36 @@
 # greenboot
-Generic Health Check Framework for systemd on [rpm-ostree](https://coreos.github.io/rpm-ostree/) based systems
+Generic Health Check Framework for systemd on [rpm-ostree](https://coreos.github.io/rpm-ostree/) based systems.
+
+## Table of contents
+* [Installation](#installation)
+* [Usage](#usage)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,47 +1,36 @@
...
 
-In order to get a full Greenboot installationon Fedora Silverblue, Fedora IoT or Fedora CoreOS:
+In order to get a full Greenboot installation on Fedora Silverblue, Fedora IoT or Fedora CoreOS:
 ```
 rpm-ostree install greenboot greenboot-default-health-checks
 
 systemctl reboot
 ```
 
-## How does it work?
```
nullr0ute, line 16: I think a how it works section makes sense, but maybe this was out of date?

jnogol: Mmm maybe incomplete because it doesn't tell how the status is outputed to the MOTD in case the checks aren't passing, but I would say the rest of the info is up to date.

jnogol: Updated the README with how the status is outputed to the MOTD in case the checks aren't passing

nullr0ute (approved):
LGTM

PR #70, opened 2022-05-31
Monitor services

say-paul:
User can select the set of services that needs to be running, default is sshd and NetworkManager.
If the selected service fail, then system will attempt a restart as per the threshold in config file by decrementing the boot_counter leading to -1, which will eventually trigger the rollback.

### DIFF
--- a/etc/greenboot/greenboot.conf
+++ b/etc/greenboot/greenboot.conf
@@ -13,3 +13,6 @@ GREENBOOT_WATCHDOG_CHECK_ENABLED=true
 ### the new deployment as culprit of reboot.
 ### It has to be a positive integer. Defaults to 24 (hours).
 # GREENBOOT_WATCHDOG_GRACE_PERIOD=24
+
+## Service monitor
+# GREENBOOT_MONITOR_SERVICES="sshd NetworkManager"
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -88,6 +88,7 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %systemd_post greenboot-grub2-set-success.service
 %systemd_post greenboot-rpm-ostree-grub2-check-fallback.service
 %systemd_post redboot-auto-reboot.service
+%systemd_post greenboot-service-monitor.service
 
 %post default-health-checks
 %systemd_post greenboot-loading-message.service
@@ -102,6 +103,7 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %systemd_preun greenboot-grub2-set-counter.service
 %systemd_preun greenboot-grub2-set-success.service
 %systemd_preun greenboot-rpm-ostree-grub2-check-fallback.service
+%systemd_preun greenboot-service-monitor.service
 
 %preun default-health-checks
 %systemd_preun greenboot-loading-message.service
@@ -116,6 +118,7 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %systemd_postun greenboot-grub2-set-counter.service
 %systemd_postun greenboot-grub2-set-success.service
 %systemd_postun greenboot-rpm-ostree-grub2-check-fallback.service
+%systemd_postun greenboot-service-monitor.service
 
 %postun default-health-checks
 %systemd_postun greenboot-loading-message.service
@@ -156,6 +159,8 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %{_unitdir}/greenboot-rpm-ostree-grub2-check-fallback.service
 %{_libexecdir}/%{name}/redboot-auto-reboot
 %{_unitdir}/redboot-auto-reboot.service
+%{_libexecdir}/%{name}/greenboot-service-monitor
+%{_unitdir}/greenboot-service-monitor.service
 
 %files default-health-checks
 %{_prefix}/lib/%{name}/check/required.d/01_repository_dns_check.sh
--- a/usr/lib/systemd/system/greenboot-service-monitor.service
+++ b/usr/lib/systemd/system/greenboot-service-monitor.service
@@ -1,0 +1,21 @@
+#  SPDX-License-Identifier: LGPL-2.1+
+#
+#  This file is part of greenboot.
+#
+#  greenboot is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=Monitor user services
+Before=boot-complete.target
+OnFailure=redboot.target
+After=default.target graphical.target multi-user.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/greenboot/greenboot-service-monitor
+
+[Install]
+RequiredBy=boot-complete.target
--- a/usr/libexec/greenboot/greenboot-service-monitor
+++ b/usr/libexec/greenboot/greenboot-service-monitor
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -eo pipefail
+GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/greenboot.conf
+service_started=true
+if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
+    source $GREENBOOT_CONFIGURATION_FILE
+fi
+
+if [ -n "$GREENBOOT_MONITOR_SERVICES" ]; then
+    services=($GREENBOOT_MONITOR_SERVICES)
+else
+    echo "GREENBOOT_MONITOR_SERVICE not found, using default value"
+    services=( sshd NetworkManager )
+fi
+
+# check all the services listed
+for service in "${services[@]}"; do
+rc="$(systemctl show -p ActiveState --value $service)"
+if [[ $rc != "active" ]]; then
+  echo "<0>${service} is ${rc}"
+  service_started=false
+fi
+done
+
+# fail if any of the monitored services failed. This will make boot-complete.target fail.
+if ! $service_started ; then
+    exit 1
+fi


say-paul:
@nullr0ute  @LorbusChris  some feedback on this will be helpful.

LorbusChris:
I only had a quick glance, but it looks to me as though you'll want to have boot-complete require the monitor, and not the other way round. That way you won't have to re-implement the reboot and set-success services in the monitor script.

say-paul:
tried this:

greenboot-service-monitor.service
`[Unit]
Description=Monitor user services
Before=boot-complete.target
OnFailure=reboot.target

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/libexec/greenboot/greenboot-service-monitor

[Install]
RequiredBy=boot-complete.target`

and 

greenboot-service-monitor as:

`#!/bin/bash
set -eo pipefail
GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/greenboot.conf
service_started=true
if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
    source $GREENBOOT_CONFIGURATION_FILE
fi

if [ -n "$GREENBOOT_MONITOR_SERVICES" ]; then
    services=($GREENBOOT_MONITOR_SERVICES)
else
    echo "GREENBOOT_MONITOR_SERVICE not found, using default value"
    services=( sshd NetworkManager )
fi

#check all the services listed
for service in "${services[@]}"; do
rc="$(systemctl show -p ActiveState --value $service)"
if [[ $rc != "active" ]]; then
  echo "${service} did not start"
  service_started=false
fi
done

#reboot if any service failed
if ! $service_started ; then
  boot_counter="$(grub2-editenv list | sed -n -E -e 's/^boot_counter=([-]?[0-9]+).*/\1/p')"
  if [ ! -n "$boot_counter" ]; then
    #  grub2-editenv - unset boot_success
    /usr/libexec/greenboot/greenboot-grub2-set-counter
    #load again to do further oprations
    boot_counter="$(grub2-editenv list | sed -n -E -e 's/^boot_counter=([-]?[0-9]+).*/\1/p')"
  fi

  if [ "$boot_counter" -gt 0 ]; then
    ((boot_counter--))
    grub2-editenv - set boot_counter="$boot_counter"
  else
    grub2-editenv - set boot_counter=-1
  fi
  exit 1
fi
`

things get into infinite restart loop , as sshd or NetworkManager dint start up.

So checked for added the service after boot-complete is reached.

LorbusChris:
have you tried something like this (we have services for setting boot_success and decrementing the counter already IIRC) :
```bash
#!/bin/bash
set -eo pipefail
GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/greenboot.conf
service_started=true
if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
    source $GREENBOOT_CONFIGURATION_FILE
fi

if [ -n "$GREENBOOT_MONITOR_SERVICES" ]; then
    services=($GREENBOOT_MONITOR_SERVICES)
else
    echo "GREENBOOT_MONITOR_SERVICE not found, using default value"
    services=( sshd NetworkManager )
fi

#check all the services listed
for service in "${services[@]}"; do
rc="$(systemctl show -p ActiveState --value $service)"
if [[ $rc != "active" ]]; then
  echo "${service} did not start"
  service_started=false
fi
done

if ! $service_started ; then
  exit 1
fi
```

say-paul:
I could be missing something , but I dont see anything where the boot_counter is decremented after a restart. Also since the check happens after boot-complete is achieved which conflicts with redboot.target, so I added the reboot. Do you see any issue with this approach?

LorbusChris:
As I mentioned, the check should run before boot-complete, effectively gating whether it can be reached or not. If that fails, redboot will be triggered automatically. The counter decrement happens in grub, if boot_success is not set, and it should not be re-implemented here. See https://github.com/rhboot/grub2/blob/fedora-37/util/grub.d/08_fallback_counting.in and https://github.com/rhboot/grub2/blob/fedora-37/util/grub.d/10_reset_boot_success.in for the relevant grub scripts.

LorbusChris:
Does that mean it's now working as intended?
You might want to throw a `Requires=network-online.target` on the unit as well to make sure this runs after the system has attempted to start NM.
And maybe a `WantedBy=multi-user.target` as well.

LorbusChris:
Also note that with having stuff in [Install], you'll have to manually enable this service until there is a preset to do this automatically.

usr/lib/systemd/system/greenboot-grub2-set-success.service
```diff
@@ -9,8 +9,8 @@
 
 [Unit]
 Description=Mark boot as successful in grubenv
-Requires=boot-complete.target
-After=boot-complete.target
+Requires=boot-complete.target greenboot-service-monitor.service
+After=boot-complete.target greenboot-service-monitor.service
```
LorbusChris, line 13: This is not needed anymore, as it's implicitly required by boot-complete now

LorbusChris:
I'm out of ideas for now. Mind pushing the updated code here for me to take a look?

usr/libexec/greenboot/greenboot-service-monitor
```diff
@@ -17,27 +17,12 @@ fi
 for service in "${services[@]}"; do
 rc="$(systemctl show -p ActiveState --value $service)"
 if [[ $rc != "active" ]]; then
-  echo "${service} did not start"
```
LorbusChris: let's record the actual state if it's not `active` (echo ${rc})

usr/lib/systemd/system/greenboot-service-monitor.service
```diff
@@ -9,12 +9,14 @@
 
 [Unit]
 Description=Monitor user services
-Requires=boot-complete.target
-After=boot-complete.target
+Before=boot-complete.target
+OnFailure=redboot.target
+Requires=network-online.target
```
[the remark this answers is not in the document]
say-paul: With this approach the service-monitor error out as sshd does not start, Probably due to the way/sequence  sshd service is initialised:
sshd.service file->
[Unit]
Description=OpenSSH server daemon
Documentation=man:sshd(8) man:sshd_config(5)
After=network.target sshd-keygen.target
Wants=sshd-keygen.target

[Service]
Type=notify
EnvironmentFile=-/etc/sysconfig/sshd
ExecStart=/usr/sbin/sshd -D $OPTIONS
ExecReload=/bin/kill -HUP $MAINPID
KillMode=process
Restart=on-failure
RestartSec=42s

[Install]
WantedBy=multi-user.target

So seems we have start the monitor-service after multi-user.target is finished.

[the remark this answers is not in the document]
LorbusChris: To me this looks like multi-user-target will try to start sshd.service. 
On `service-monitor-service`, try adding this:
```
After=network.target sshd-keygen.target
Requires=network.target sshd-keygen.target
```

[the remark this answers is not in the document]
say-paul: That did not work, It was one of my initial attempt to able to check status and to write minimum lines of code.
I am not sure how we can check for the service when it is wanted by the end state(multi-user.target), which comes after boot-complete.target

say-paul:
reverting back to working script code. 
@LorbusChris @runcom  thoughts on how to take it forward?

LorbusChris:
I think we were pretty close before and only needed to get the sorting of the involved service starts right, but this re-revert isn't making things easier to review (I'd suggest you revert the revert, and squash all changes). Did you check what the status of sshd service actually is? (https://github.com/fedora-iot/greenboot/pull/70#discussion_r894529615) Maybe that'll give us a pointer.

usr/libexec/greenboot/greenboot-service-monitor
```diff
@@ -0,0 +1,28 @@
...
+else
+    echo "GREENBOOT_MONITOR_SERVICE not found, using default value"
+    services=( sshd NetworkManager )
+fi
+
+#check all the services listed
+for service in "${services[@]}"; do
+rc="$(systemctl show -p ActiveState --value $service)"
+if [[ $rc != "active" ]]; then
+  echo "<0>${service} did not start"
```
LorbusChris, line 20: please echo the actual ActiveState here. It can be either `active`, `inactive`, `activating`, `deactivating`, or `failed`.

LorbusChris:
@say-paul let's move this forward :)
What state did sshd report in the end? We still need to order the service start correctly for it to work (i.e. the monitor should start after sshd). Probably after multi-user.target after all.
There is a similar service included in systemd, which checks whether any service has failed, and which also gates boot-complete.target:
- https://www.freedesktop.org/software/systemd/man/systemd-boot-check-no-failures.service.html
- https://github.com/systemd/systemd/blob/main/units/systemd-boot-check-no-failures.service.in
 You probably want to lift the following from that unit for the monitor:
```
After=default.target graphical.target multi-user.target
Before=boot-complete.target
```
Please squash all changes into one commit and report the progress back here.

usr/lib/systemd/system/greenboot-service-monitor.service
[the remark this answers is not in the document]
LorbusChris: multi-user.target should come before boot-complete.target

say-paul:
Adding the things similar to system-boot-check-no-failure.service i was facing `Freezing execution` and the system didi not boot, however after adding `After=network.target sshd.service` I was able to boot successfully.

usr/lib/systemd/system/greenboot-service-monitor.service
```diff
@@ -0,0 +1,22 @@
...
+#  greenboot is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=Monitor user services
+Before=boot-complete.target
+OnFailure=redboot.target
+After=network.target sshd.service
```
LorbusChris, line 14: Could you pull in the change from https://github.com/fedora-iot/greenboot/pull/72 and try this?
```suggestion
After=default.target graphical.target multi-user.target
```

LorbusChris (changes requested):
See suggestions above and below. Please also provide a proper commit message for this change

usr/lib/systemd/system/greenboot-service-monitor.service
```diff
@@ -0,0 +1,22 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/greenboot/greenboot-service-monitor
+
+[Install]
+RequiredBy=boot-complete.target
+WantedBy=multi-user.target
```
LorbusChris, line 22: this last line also shouldn't be needed

etc/greenboot/greenboot.conf
```diff
@@ -7,9 +7,12 @@
 ### This variable controls 
 ### This value can be "true, TRUE, True..." as it will be lowercased.
 ### Set it to anything else to disable this check.
-GREENBOOT_WATCHDOG_CHECK_ENABLED=true
+# GREENBOOT_WATCHDOG_CHECK_ENABLED=true
```
LorbusChris, line 10: is this disabled for testing purposes?

usr/lib/systemd/system/greenboot-service-monitor.service
say-paul: do this causing `systemd: Freezing execution ` on reboot, similar to when trying to enable system-boot-check-no-failure.service, not sure if its my system or things in general.

say-paul: @LorbusChris could you please test your suggested method and let me know?

LorbusChris: I don't see the commit from #72 in this PR. Did you try to test this together with that change? It is crucial that is added here. You can simply cherry-pick it on top of this branch. Please provide full output logs of any failures, I don't have cycles to test this.

say-paul:
@LorbusChris  its working now .

greenboot.spec
```diff
@@ -156,6 +160,8 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %{_unitdir}/greenboot-rpm-ostree-grub2-check-fallback.service
 %{_libexecdir}/%{name}/redboot-auto-reboot
 %{_unitdir}/redboot-auto-reboot.service
+%{_unitdir}/greenboot-service-monitor.service
+%{_libexecdir}/%{name}/greenboot-service-monitor
```
LorbusChris, line 164:
```suggestion
%{_libexecdir}/%{name}/greenboot-service-monitor
%{_unitdir}/greenboot-service-monitor.service
```

greenboot.spec
```diff
@@ -88,6 +88,8 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %systemd_post greenboot-grub2-set-success.service
 %systemd_post greenboot-rpm-ostree-grub2-check-fallback.service
 %systemd_post redboot-auto-reboot.service
+%systemd_post greenboot-service-monitor.service
+systemctl enable greenboot-service-monitor.service
```
LorbusChris, line 92:
```suggestion
```
This has to be dropped here, and instead has to be done via a preset

usr/lib/systemd/system/greenboot-grub2-set-success.service
```diff
@@ -19,4 +19,4 @@ ExecStart=/usr/bin/grub2-editenv - set boot_success=1
 ExecStart=/usr/bin/grub2-editenv - unset boot_counter
 
 [Install]
-RequiredBy=multi-user.target
+WantedBy=multi-user.target
```
LorbusChris, line 22: please drop this change from this commit and instead cherry-pick the commit from https://github.com/fedora-iot/greenboot/pull/72 onto this branch (it has a commit message why this change is needed)

say-paul: It was cherry-picked but squashing commits removes other messages, so dropping it from here and rebasing from main where #72  which is already merged.

LorbusChris (changes requested):
some more nits. and regarding the commit message, maybe you could formulate it like this:

```
Add greenboot-service-monitor.service for service health checking

This change adds a service to monitor whether or not specified services have started correctly.
If enabled, the service runs after `multi-user.target` is reached, and is required to reach `boot-complete.target`.
The monitored services can be configured in `greenboot.conf`. By default `sshd` and `NetworkManager` are monitored.

```

LorbusChris:
This LGTM now. Thanks for working on this @say-paul.

LorbusChris:
@say-paul as a follow-up, please make sure to create a preset for enabling this service (probably needs to land in both fedora iot and well as rhel for edge)

etc/greenboot/greenboot.conf
runcom: this was kept like this, and now it's regressing in CI

usr/libexec/greenboot/greenboot-service-monitor
```diff
@@ -0,0 +1,28 @@
...
+service_started=true
+if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
+    source $GREENBOOT_CONFIGURATION_FILE
+fi
+
+if [ -n "$GREENBOOT_MONITOR_SERVICES" ]; then
+    services=($GREENBOOT_MONITOR_SERVICES)
+else
+    echo "GREENBOOT_MONITOR_SERVICE not found, using default value"
+    services=( sshd NetworkManager )
```
runcom, line 13: I still think we have to drop this, and then any distribution can just set a default in the /etc config they ship ;)

LorbusChris: +1 to just setting the default services in the config that's shipped instead of hardcoding it here

say-paul:
> @say-paul as a follow-up, please make sure to create a preset for enabling this service (probably needs to land in both fedora iot and well as rhel for edge)

its already there for rhel

PR #72, opened 2022-07-06
set-success.service: Change RequiredBy to WantedBy=multi-user.target

LorbusChris:
This change weakens the dependency of `multi-user.target` on the
set-success service. This is so that `multi-user.target` can be reached
before `boot-complete.target`, for which this service is a requirement.
This new behavior will enable services such as systemd's
`systemd-boot-check-no-failures.service`, which runs after
`multi-user.target` but still gates whether `boot-complete.target` can
be reached or not.


say-paul:
Looks good to me, works well with pr [70](https://github.com/fedora-iot/greenboot/pull/70)

PR #73, opened 2022-07-19
release 0.15.0

say-paul:
Version bump due to new feature

### DIFF
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -1,7 +1,7 @@
 %global debug_package %{nil}
 
 Name:               greenboot
-Version:            0.14.0
+Version:            0.15.0
 Release:            1%{?dist}
 Summary:            Generic Health Check Framework for systemd
 License:            LGPLv2+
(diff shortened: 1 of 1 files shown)


greenboot.spec
```diff
@@ -169,6 +169,10 @@ install -DpZm 0755 usr/lib/greenboot/check/wanted.d/* %{buildroot}%{_prefix}/lib
 %{_prefix}/lib/%{name}/check/required.d/02_watchdog.sh
 
 %changelog
+* Thu Jul 15 2022 Sayan Paul <paul.sayan@gmail.com> - 0.15.0-1
```
LorbusChris, line 172: This date is not correct, it's either Fri 15, or Thu 14, but this should now become today's date.

LorbusChris (changes requested):

greenboot.spec
say-paul: I need to learn to see the calendar again!!!

PR #74, opened 2022-08-02
spec update to add conf during installation

say-paul:
this fix will add greenboot.conf in /etc/greenboot/ folder
this is required to enbale config management via osbuild


say-paul:
@runcom @nullr0ute @LorbusChris  need some eyes.

runcom:
ah nice! :+1:

nullr0ute:
LGTM

PR #77, opened 2022-08-29
Fix ordering cycles starting with multi-user.target

LorbusChris:
This change introduces a new greenboot-success.target which requires
boot-complete.target and multi-user.target.
This is to avoid an ordering cycle that was introduced with the recent
addition of the service monitor, which explicitly orders
multi-user.target before boot-complete.target.
The start of services that were either WantedBy= or
RequiredBy=multi-user.target but which also had to run
After=boot-complete.target, couldn't be ordered properly
anymore. This change fixes that.

An unfortunate side-effect of this is that `multi-user.target` can't be
used as the default target anymore. The default target now has to be
changed to `greenboot-success.target` for greenboot to run.

---

If we go ahead with this approach, the `default_target` in the Fedora IoT/RHEL for Edge
rpm-ostree manifests will have to be changed to `greenboot-success.target`.
For Fedora IoT, the default target is specified here: https://pagure.io/fedora-iot/ostree/blob/main/f/fedora-iot-base.json#_126

@runcom @say-paul

### DIFF
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -90,6 +90,7 @@ install -DpZm 0644 etc/greenboot/greenboot.conf %{buildroot}%{_sysconfdir}/%{nam
 %systemd_post greenboot-rpm-ostree-grub2-check-fallback.service
 %systemd_post redboot-auto-reboot.service
 %systemd_post greenboot-service-monitor.service
+%systemd_post greenboot-success.target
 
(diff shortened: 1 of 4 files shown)


LorbusChris:
Closing in favor of https://github.com/fedora-iot/greenboot/pull/78

PR #78, opened 2022-08-29
disable DefaultDependencies to fix cycle error

say-paul:
The fix relaxes the dependecies of task-runner and status service that
otherwise causes issue in the service to start as well as fix the missing
logs
signed off: saypaul@fedoraproject.org


LorbusChris:
Makes sense. LGTM

LorbusChris (approved):

PR #79, opened 2022-09-08
avoid running checks if conditions aren't met

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

### DIFF
--- a/usr/lib/greenboot/check/required.d/02_watchdog.sh
+++ b/usr/lib/greenboot/check/required.d/02_watchdog.sh
@@ -41,6 +41,11 @@ check_if_current_boot_is_wd_triggered() {
     fi
 }
 
+if ! check_if_there_is_a_watchdog ; then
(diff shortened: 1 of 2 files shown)


usr/lib/greenboot/check/required.d/02_watchdog.sh
```diff
@@ -41,6 +41,11 @@ check_if_current_boot_is_wd_triggered() {
     fi
 }
 
+if !check_if_there_is_a_watchdog ; then
```
miabbott, line 44:
```$ shellcheck usr/lib/greenboot/check/required.d/02_watchdog.sh

...

In usr/lib/greenboot/check/required.d/02_watchdog.sh line 44:
if !check_if_there_is_a_watchdog ; then
    ^-- SC1035: You are missing a required space after the !.
```

PR #80, opened 2022-09-08
add ShellCheck CI + update code accordingly

miabbott:
This is a draft PR to add ShellCheck CI to the GitHub Actions configuration and updates the shell scripts according to the feedback from ShellCheck.

I've not tested this on a real system, so the changes might actually break things.

I'm putting this up here as an example of how we can expand CI here. I would love to see someone pick up the changes and drive them to completion.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 name: Run CI tests
-on: 
+on:
   - push
   - pull_request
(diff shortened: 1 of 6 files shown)


paulwhalen (commented):
Changes LGTM.

paulwhalen (approved):
LGTM

PR #83, opened 2022-09-11
Added additional check to skip disabled services

say-paul:
Service monitor is failing if the services are disabled. The fix adds a layer of check to guarantee the services are enabled to avoid any false negative.

Signed-off-by: Sayan Paul <paul.sayan@gmail.com>

### DIFF
--- a/usr/libexec/greenboot/greenboot-service-monitor
+++ b/usr/libexec/greenboot/greenboot-service-monitor
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -o pipefail
+set -eo pipefail
 GREENBOOT_CONFIGURATION_FILE=/etc/greenboot/greenboot.conf
 service_started=true
 if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
@@ -15,17 +15,17 @@ fi
 
 # check all the services listed
 for service in "${services[@]}"; do
-state="$(systemctl is-enabled $service)"
-if [[ $state == "enabled" ]]; then
+if [[ $(systemctl is-enabled $service) == "enabled" ]]; then
     rc="$(systemctl show -p ActiveState --value $service)"
     if [[ $rc != "active" ]]; then
         echo "<0>${service} is ${rc}"
         service_started=false
     fi
 else
-    echo "Skipped $state service: $service"
+    echo "Skipped service: $service"
 fi
 done
+
 # fail if any of the monitored services failed. This will make boot-complete.target fail.
 if ! $service_started ; then
     exit 1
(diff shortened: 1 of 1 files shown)


say-paul:
@LorbusChris Additional check has been placed, though monitor needs the ability to set the counter incase any service fails to start.

LorbusChris:
I don't think the monitor has to do that, since the monitor runs after a boot, not after an upgrade/before reboot.
If anything, maybe there should be another conditional for the monitor to only run when the boot_counter is set.

Changes here LGTM, but I'll defer to @runcom

usr/libexec/greenboot/greenboot-service-monitor
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eo pipefail
+set -o pipefail
```
runcom, line 2: why this? :thinking:

runcom:
is this making sure that if somebody disable a service which is monitored we can still boot and everything? will this cover what qe tested?

usr/libexec/greenboot/greenboot-service-monitor
say-paul: `state="$(systemctl is-enabled $service)"` exits>0 if any service is disabled, so removed -e.

runcom: if you put the subshell `$(..)` in the if statement it shoulnd' fail right?

say-paul:
@runcom yes, I have written a test service too which test the restart. but still trying to figure out how to test the rollback.

say-paul:
@LorbusChris but the purpose of the service monitor is not fulfilled, when a service fails , it still shows: SYSTEM is UNHEALTHY, but boot_counter is unset in grubenv. Manual intervention necessary. 
the system needs to restart on failure.

LorbusChris:
greenboot isn't really supposed to do anything in the case of a simple reboot, because we don't want to roll back when the current OS version has already shown success in an earlier boot.

If the monitor doesn't run (or just exits 0) in case of simple reboot without boot_counter, nothing should show a failure in the MOTD, no?

LorbusChris:
> is this making sure that if somebody disable a service which is monitored we can still boot and everything? will this cover what qe tested?

yes, this is what this change should ensure

say-paul:
@LorbusChris seems the notion of having service monitor is to only monitor things post upgrade which absolutely makes sense.
@runcom @miabbott Do we need to explicitly mention it somewhere as I am not aware of this earlier, so can be the others.

usr/libexec/greenboot/greenboot-service-monitor
say-paul: yes, that worked, updated the script.

nullr0ute:
I don't think this is needed any more @say-paul

say-paul:
@nullr0ute not required anymore, closing it.

PR #84, opened 2022-09-22
greenboot-grub2-set-counter.service: ensure /boot is mounted

AdamWill:
See https://bugzilla.redhat.com/show_bug.cgi?id=2121944 . As it stands, this service can actually race with the mounting of /boot, and fail if it starts before /boot is mounted. Adding this ensures that doesn't happen.

Signed-off-by: Adam Williamson <awilliam@redhat.com>


nullr0ute (approved):
LGTM

AdamWill:
The CI fail doesn't look like it's related.

LorbusChris:
This would just fail in case boot is not yet mounted no?
Shouldn't we instead do `After=boot.mount` and `Requires=boot.mount`?

LorbusChris (approved):

paulwhalen (commented):
LGTM

PR #85, opened 2022-10-19
Revert service monitor

runcom:
The feature is half backed anyway and it'll cause more harm than good

revert https://github.com/fedora-iot/greenboot/pull/70
revert https://github.com/fedora-iot/greenboot/pull/78

I'm pretty sure #78 was a fallout from #70, hence why I'm reverting it

@LorbusChris ptal

cc @nullr0ute 

we'll need a new release once this is merged, let's keep greenboot working the way it was before w/o adding additional things which only cause bugs at worst


LorbusChris (approved):
I think the `DefaultDependencies=No` could stay, but reverting it won't hurt either

LGTM

Issue #93, opened 2023-01-20
Failing 01_update_platforms_check.sh on Fedora IoT


alcir:
I don't know if it is a duplicate of #90 and #71, so excuse me in advance:

On a fresh Fedora IoT 37 I always get 

```text
There are problems connecting with the following URLs:
https://ostree.fedoraproject.org/iot
https://ostree.fedoraproject.org/iot/mirrorlist
```

Also running `/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh` by hand after the boot, it returns the same result.

I'm not a great bash expert, but I suspect that the double quotes around the `${UPDATE_PLATFORM_URLS[@]}` will result in a single line containing all the UPDATE_PLATFORM_URLS making ineffective the for loop
https://github.com/fedora-iot/greenboot/blob/7ad3ea53fd4154005bb26d2d48d41d4dc42853e1/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh#L16

At the end of the day, the curl command is
`curl -o /dev/null -Isw '%{http_code}\n' "https://ostree.fedoraproject.org/iot https://ostree.fedoraproject.org/iot/mirrorlist"`
leading to a 000 HTTP_STATUS variable (that is curl: (3) URL using bad/illegal format or missing URL)

In short: if in such line I remove the double quotes
`for UPDATE_PLATFORM_URL in ${UPDATE_PLATFORM_URLS[@]}; do`
the script works as expected.

paulwhalen:
This regression happened recently with https://github.com/fedora-iot/greenboot/commit/84bbd674a7da12ea7c07657129d489942c69e449

paulwhalen:
@alcir did you want to open a RHBZ for this issue so we can track it for F38? (if not I can)

Thanks for finding it!

alcir:
> @alcir did you want to open a RHBZ for this issue so we can track it for F38? (if not I can)

I will

alcir:
It is already tracked
https://bugzilla.redhat.com/show_bug.cgi?id=2062455

alcir:
Again, I'm not a bash expert, but instead of disable shellcheck, here 
https://github.com/fedora-iot/greenboot/blob/7ad3ea53fd4154005bb26d2d48d41d4dc42853e1/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh#L8
we could actually populate the UPDATE_PLATFORM_URLS array simply in this way
[...]

PR #94, opened 2023-01-20
disable shellcheck SC2048 on URL array

miabbott:
See: https://github.com/koalaman/shellcheck/wiki/SC2048

In the case of the UPDATE_PLATFORM_URLS variable, we actually want to
split on whitespace since the array should just contain URLs.

Disable the Shellcheck directive for this array.

Closes: https://github.com/fedora-iot/greenboot/issues/93


miabbott:
I haven't tested this (which got us in trouble the first time), so let's hold off on merging until we have confidence in this change.

paulwhalen (approved):
LGTM

alcir:
I'm not a bash expert, but instead of disable shellcheck, here

https://github.com/fedora-iot/greenboot/blob/7ad3ea53fd4154005bb26d2d48d41d4dc42853e1/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh#L8

we could actually populate the `UPDATE_PLATFORM_URLS` array simply in this way
```
UPDATE_PLATFORM_URLS=($(grep -P -ho 'http[s]?.*' $REPOS_DIRECTORY/*))
```
or
```
readarray -t UPDATE_PLATFORM_URLS < <(grep -P -ho 'http[s]?.*' $REPOS_DIRECTORY/*)
```

miabbott:
> I'm not a bash expert, but instead of disable shellcheck, here
> 
> https://github.com/fedora-iot/greenboot/blob/7ad3ea53fd4154005bb26d2d48d41d4dc42853e1/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh#L8
> 
> we could actually populate the `UPDATE_PLATFORM_URLS` array simply in this way
> 
> ```
> UPDATE_PLATFORM_URLS=($(grep -P -ho 'http[s]?.*' $REPOS_DIRECTORY/*))
> ```
> 
> or
> 
> ```
> readarray -t UPDATE_PLATFORM_URLS < <(grep -P -ho 'http[s]?.*' $REPOS_DIRECTORY/*)
> ```

I tried both methods of assigning the array; the first suggestion introduced an error with Shellcheck related to [SC20227](https://www.shellcheck.net/wiki/SC2207) and the second suggestion did not trigger any error.

However, if I remove the comment to disable the SC2048 check on [line 16-17](https://github.com/miabbott/greenboot/blob/array_fix/usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh#L16-L17), the error returns regardless of how the array is originally assigned.

alcir:
What about 
```
8     mapfile -t UPDATE_PLATFORM_URLS < <(grep -P -ho 'http[s]?.*' "${REPOS_DIRECTORY}"/*)
9     if [[ ${#UPDATE_PLATFORM_URLS[@]} -eq 0 ]]; then
...
16     for UPDATE_PLATFORM_URL in "${UPDATE_PLATFORM_URLS[@]}"; do
```

Running `shellcheck` from command line doesn't return any error and the script works as intended.

miabbott:
> 

This looks promising...I was heading in this direction but lack of time has kept me from pursuing this fully.

nullr0ute:
@miabbott so the bats_test is failing, is that a problem or does the test need to be adjusted or is it unrelated to this? Other than that is this ready to review/merge?

miabbott:
@nullr0ute BATS is failing like this:

```
not ok 1 Ensure watchdog check is working
# (from function `source' in file /usr/lib/greenboot/check/required.d/02_watchdog.sh, line 47,
#  from function `setup' in test file check_watchdog_support.bats, line 4)
#   `source $GREENBOOT_DEFAULT_CHECK_PATH/required.d/02_watchdog.sh --source-only' failed
# No watchdog on the system, skipping check
```

Looking at the history of that test on other PRs, it is failing similarly, so I believe it is unrelated to this change.

I believe this is ready for review + merge

PR #97, opened 2023-02-22
Run scripts under /usr/lib/greenboot

eric-ch:
This is an RFC as I may be misunderstanding the file-hierarchy, but I thought it would be clearer to ask with these associated changes. I used [systemd file-hierarchy](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#/usr/lib/) as a reference point. The documentation of {greenboot} did not mention `/usr/lib/greenboot/{green,red}.d`, yet the specfile lists them.

The specfile creates:
 * `/usr/lib/greenboot/green.d`
 * `/usr/lib/greenboot/red.d`

Yet "/usr/libexec/greenboot/greenboot [green|red]" will not run the scripts in these directories. Add and run them before the ones in /etc.


miabbott (changes requested):
This is a good catch!

Since `/usr/lib` is populated by the vendor (i.e. distribution), it is possible that when `greenboot` is packaged, the vendor would want to stick particular red/green scripts in that location.  So we definitely want to support this use case.

I'd request that some additional unit tests are added to check this functionality.

say-paul:
LGTM , I agree with @miabbott to add some test for this.

eric-ch:
@miabbott, @say-paul thanks for the review!

Since some unit-tests already checked for vendor and system specific scripts (`/usr/lib` vs `/etc`) I tried extending that. I am happy to revisit this if that is too intrusive.

Also I noticed that one unit test mentioned that all scripts under `{/usr/lib,/etc}/greenboot/check/required.d` should run, even if one failed. It appeared that was not tested for all the directory hierarchies and greenboot would skip the scripts under `/etc` if any  under `/usr/lib` failed. The last commit attends that, taking the description in the unit test as the reference.

say-paul:
@eric-ch Nice Catch, Tests looks good to me.

ericcurtin:
LGTM

eric-ch:
Hi @say-paul , @miabbott, gentle ping unless this has fallen out of favor. Is there anything I can do to help moving this along?

say-paul:
cc @nullr0ute

nullr0ute (approved):
LGTM

PR #106, opened 2023-06-28
Use differential-shellcheck instead of action-shellcheck

jamacku:
This change doesn't change the core behavior of ShellCheck linting. Differential ShellCheck uses by default [flag `-X`](https://github.com/redhat-plumbers-in-action/differential-shellcheck#external-sources). It will also [exclude](https://github.com/redhat-plumbers-in-action/differential-shellcheck#exclude-path) all shell scripts located in `tests/` from linting.

#### Some benefits of using differential ShellCheck Action

Differential ShellCheck is a GitHub Action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

It is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,11 +14,26 @@ jobs:
   shellcheck:
     name: Shellcheck
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      security-events: write
+
     steps:
-      - uses: actions/checkout@v2
-      - name: Run Shellcheck
-        uses: ludeeus/action-shellcheck@master
-        env:
-          SHELLCHECK_OPTS: -x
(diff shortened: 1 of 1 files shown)


say-paul:
This looks cool!! do you have its as IDE-plugin(vs code) as static analysis tool?

nullr0ute:
LGTM

Issue #98, opened 2023-03-03
RFE: make `01_update_platforms_check.sh` more intelligent/robust


miabbott:
An ostree remote config may have a `url=` parameter and a `contenturl=` parameter included in the config. When the `contenturl=` parameter is present, the ostree client will fetch *content* from that resource, but will fetch *metadata* from the resource specified by the `url=` parameter.

Currently, the Fedora IoT ostree infrastructure is configured in a way that doing `curl -L https://ostree.fedoraproject.org/iot` (as specified in the `url=` parameter) returns an HTTP 403.  But `curl -L https://ostree.fedoraproject.org/iot/config` returns HTTP 200.

Along similar lines, `curl -L https://ostree.fedoraproject.org/iot/mirrorlist` returns HTTP 200. And substituting the CloudFront hostname from the mirror list: `curl -L https://d2ju0wfl996cmc.cloudfront.net/config` also returns HTTP 200.

The intent is to make the script more intelligent to test for actual content availability depending on how the ostree remote config is populated.  If there is a `contenturl=` parameter, the script should check that fetching `config` asset from both the `url=` and `contenturl=` parameters to validate both more completely. In the absence of the `contenturl=` parameter, the script should only check the `url=` parameter.

See also: https://issues.redhat.com/browse/THEEDGE-3108

miabbott:
See also #71

runcom:
I think all these are valid, easy additions that we can make (easily) and since the rewrite will support bare bash scripts like this I think it's worth doing

miabbott:
https://issues.redhat.com/browse/THEEDGE-3108

nullr0ute:
> Another complexity to be aware of: some Red Hat Edge systems will have the update URLs protected by entitlement certificates, so bare `curl`'ing of the the URLs will always fail.
> 
> Any `curl` operations for those kinds of URLs will need to make use of `curl --cacert <RH CA cert> --cert <client entitlement cert> --key <client entitlement key> ...`

The end point should still be reachable even if auth fails.

miabbott:
> The end point should still be reachable even if auth fails. The idea of the test was to ensure the network stack was working, routing, DNS, HTTP etc.

We need to define "reachable" in this case.

If we `curl` the endpoint and we get a TCP connection established, does that meet the qualifications of reachable?

Or do we want to say that reachable == HTTP 200?

[...]

PR #116, opened 2023-09-21
use ostree commands to fetch remote URLs

sebhoss:
Newer fedora-iot versions don't seem to include any files in `/etc/ostree/remotes.d` anymore, yet `ostree remote list` returns a list of remotes. Therefore, we are using that and its sibling `ostree remote show-url` to get all update URLs.

Relates to #98


usr/lib/greenboot/check/wanted.d/01_update_platforms_check.sh
```diff
@@ -5,7 +5,7 @@ REPOS_DIRECTORY=/etc/ostree/remotes.d
 URLS_WITH_PROBLEMS=()
 
 get_update_platform_urls() {
-    mapfile -t UPDATE_PLATFORM_URLS < <(grep -P -ho 'http[s]?.*' "${REPOS_DIRECTORY}"/*)
+    mapfile -t UPDATE_PLATFORM_URLS < <(ostree remote list | xargs ostree remote show-url)
```
say-paul, line 8: fetching it from rpm-ostree directly makes sense but this actions needs to iterative as rpm-ostree can have multiple remotes that it can rebase off  [reference doc](https://redhat-internal.slack.com/archives/C999USB0D/p1695815913023619?thread_ts=1695815581.270249&cid=C999USB0D)

PR #126, opened 2023-10-17
ci: add greenboot-rs ci workflow

yih-redhat:


nullr0ute:
This needs to me a PR against the greenboot-rs branch, not main

yih-redhat:
The design is to enable the CI workflow in greenboot repo, but will only run for any pull request to branch greenboot-rs.
https://github.com/fedora-iot/greenboot/pull/126/files#diff-11b9ad7b2fa47d2680fcd076bb109a944e94a0691a84a53e6b29189a59ccc993R96

Never add ci workflow to a branch before(not sure if it works), I will check this.

yih-redhat:
From my understanding, the workflow (trigger by pull_request) should be put in main branch, and it can decide to run for which sub branch. In this case, the workflow will only run for pull_request that targeting greenboot-rs branch, https://github.com/fedora-iot/greenboot/pull/126/files#diff-11b9ad7b2fa47d2680fcd076bb109a944e94a0691a84a53e6b29189a59ccc993R96

For workflow that only exist in non-main branch, I googled it and found some articles that said it should exist in main to be triggered. So I made some test, added workflow in a sub-branch, and then made a pull_request to that sub-branch, the workflow cannot be triggered.

@henrywang, could you please take a look of this? as you have rich experience in github workflow.

PR #129, opened 2023-10-23
Add static grub config

cgwalters:
This pairs with https://github.com/coreos/bootupd/pull/543 to have bootupd support injecting the greenboot grub logic only if greenboot is installed.

### DIFF
--- a/greenboot.spec
+++ b/greenboot.spec
@@ -19,6 +19,7 @@ BuildRequires:      systemd-rpm-macros
 Requires:           systemd >= 240
 Requires:           grub2-tools-minimal
 Requires:           rpm-ostree
+Requires:           bootupd
 # PAM is required to programatically read motd messages from /etc/motd.d/*
 # This causes issues with RHEL-8 as the fix isn't there an el8 is on pam-1.3.x
 Requires:           pam >= 1.4.0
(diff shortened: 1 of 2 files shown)


nullr0ute:
How does this impact users that may not have moved to bootupd/unified core yet?

LorbusChris:
To add some more context: This config snippet originates with [grub2-tools](https://github.com/rhboot/grub2/blob/fedora-39/util/grub.d), but it's [excluded in FCOS](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/manifests/grub2-removals.yaml). 
I'm not familiar with the reasons of osbuild carrying another copy of that snippet.

Desktop variants currently use:
- [10_reset_boot_success](https://github.com/rhboot/grub2/blob/fedora-39/util/grub.d/10_reset_boot_success.in)
- [12_menu_auto_hide](https://github.com/rhboot/grub2/blob/fedora-39/util/grub.d/12_menu_auto_hide.in)
- [14_menu_show_once](https://github.com/rhboot/grub2/blob/fedora-39/util/grub.d/14_menu_show_once.in)
(It looks like there's [an effort to drop 12_menu_auto_hide eventually](https://github.com/rhboot/grub2/commit/39698498b51d9f4486add7f43be1a059a96bcb0a) in favor of 14_menu_show_once)

How would these snippets be shipped for e.g. the Silverblue ostree native container image? Do all the snippets have to shipped this way going forward in order to work with bootc?

Should [08_fallback_counting](https://github.com/rhboot/grub2/blob/fedora-39/util/grub.d/08_fallback_counting.in) be removed from `grub2-tools` after this merges to make this repo here the canonical location?

cgwalters:
cc https://github.com/ostreedev/ostree/pull/3162#issuecomment-1934679505

runcom:
Given low impact as per https://github.com/fedora-iot/greenboot/pull/129#issuecomment-1779710159 and since we're moving to bootupd, this ease with testing that flow too.

lgtm

PR #100, opened 2023-03-29
Simplified greenboot

say-paul:
This will implement the the existing greenboot health check functionality:
Below are the functionality that will be implemented as part of this PR:
- [ ] greenboot params set up- health-check, rollback(not implemented yet)
- [ ] Healthcheck process - required.d,wanted.d,red.d,green.d
- [ ] Working with grub params - boot_counter,boot_success
- [ ] MOTD
- [ ] log

PR #133, opened 2023-11-14
test: add test code in greenboot-rs branch

henrywang:
Let's move test code into `greenboot-rs` branch. So if any code change, test should be changed at the same PR.


say-paul:
dependent on pr #100

PR #134, opened 2023-11-14
ci: don't fetch test code from Yi's repo

henrywang:
test: remove un-used code for cleanup


yih-redhat:
The original plan is to add the test scripts in branch greenboot-rs, and then Sayan can make a rebase in his pr to trigger the test.
After discussed with Sayan, he mentioned rebase could cause some problems now, so I put the test scripts in my repo, but will move them into greenboot-rs branch after Sayan merged his pr https://github.com/fedora-iot/greenboot/pull/100

henrywang:
You can add test code into PR #100. That can avoid issue. So let's land this PR first.

henrywang:
@say-paul Could you please review this PR? Thanks!

say-paul:
I suggested against it earlier as the tests were more inline with the current green-boot, which would fail the new -rs branch. Since its now inline with the greenboot-rs, its safe to merge.

say-paul:
greenboot-rs built process is itself broken, and needs some update, merging a tests here wont give us any advantage. so we can keep the pr #133  open until #100  is merged.

But this PR makes sense and look good to me.

say-paul (approved):

Issue #119, opened 2023-09-27
provide mechanism for disabling certain default health checks


miabbott:
The opinion around #117 is that removal of the health checks is too coarse and severely limits the usefulness of `greenboot` as a whole.

Instead, let's investigate a means of enabling/disabling the default health checks.

7flying:
- we need to give the user the ability to choose whether they want the default health checks enabled or not (all of them, some of them etc), given that they may have the default-health-checks RPM installed

PR #137, opened 2024-04-03
greenboot: add feature to disable healthchecks

djach7:
Addresses #119. Allows users to specify healthchecks to be skipped via a DISABLED_HEALTHCHECKS variable in greenboot.conf. Skipped healthchecks will be reflected with an appropriate message in the logs.

### DIFF
--- a/etc/greenboot/greenboot.conf
+++ b/etc/greenboot/greenboot.conf
@@ -13,3 +13,15 @@ GREENBOOT_WATCHDOG_CHECK_ENABLED=true
 ### the new deployment as culprit of reboot.
 ### It has to be a positive integer. Defaults to 24 (hours).
 # GREENBOOT_WATCHDOG_GRACE_PERIOD=24
+
+### This variable allows you to specify healthchecks to be skipped. 
+### The healthcheck must be specified by script name, as in the 
+### example below. Multiple healthchecks may be skipped by separating
+### the script names with spaces.
+### NOTE: Script names must be spelled EXACTLY. Typos will result in
+### unwanted behaviour.
+### DISABLED_HEALTHCHECKS=(
+### 	"01_repository_dns_check.sh" 
+### 	"02_watchdog.sh"
+### )
+DISABLED_HEALTHCHECKS=()
--- a/usr/libexec/greenboot/greenboot
+++ b/usr/libexec/greenboot/greenboot
@@ -7,6 +7,23 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
 SCRIPTS_GREEN_PATHS=("/usr/lib/greenboot/green.d" "/etc/greenboot/green.d")
 SCRIPTS_RED_PATHS=("/usr/lib/greenboot/red.d" "/etc/greenboot/red.d")
 
+source_configuration_file() {
+    greenboot_configuration_file=/etc/greenboot/greenboot.conf
+    if test -f "$greenboot_configuration_file"; then
+	# shellcheck source=/etc/greenboot/greenboot.conf
+        source $greenboot_configuration_file
+    fi
+}
+
+source_configuration_file
+function is_disabled {
+    healthcheck=$1
+    for disabled_healthcheck in "${DISABLED_HEALTHCHECKS[@]}"; do
+        if [ "${healthcheck}" == "${disabled_healthcheck}" ]; then return 0; fi
+    done
+    return 1
+}
+
 script_runner () {
   local scripts_dir=$1; shift
   local mode=$1; shift
@@ -14,21 +31,25 @@ script_runner () {
   local required_hc_failed=false
   echo "$start_msg"
   for script in $(find "$scripts_dir" -name '*.sh' | sort); do
-    local rc=0
-    systemd-cat -t "$(basename "$script")" bash "$script" || rc=$?
-    if [ $rc -ne 0 ]; then
-      local failure_msg
-      failure_msg="Script '$(basename "$script")' FAILURE (exit code '$rc')"
-      case "$mode" in
-        "relaxed")
-          echo "<2>$failure_msg. Continuing..." >&2
-          ;;
-        "strict")
-          required_hc_failed=true
-          echo "<0>$failure_msg. Continuing..." >&2
-      esac
+    if is_disabled "$(basename "$script")"; then
+      echo "'$(basename "$script")' was skipped, as specified in config"
     else
-      echo "<6>Script '$(basename "$script")' SUCCESS"
+      local rc=0
+      systemd-cat -t "$(basename "$script")" bash "$script" || rc=$?
+      if [ $rc -ne 0 ]; then
+        local failure_msg
+        failure_msg="Script '$(basename "$script")' FAILURE (exit code '$rc')"
+        case "$mode" in
+          "relaxed")
+            echo "<2>$failure_msg. Continuing..." >&2
+            ;;
+          "strict")
+            required_hc_failed=true
+            echo "<0>$failure_msg. Continuing..." >&2
+        esac
+      else
+        echo "<6>Script '$(basename "$script")' SUCCESS"
+      fi
     fi
   done
 


etc/greenboot/greenboot.conf
```diff
@@ -12,4 +12,11 @@ GREENBOOT_WATCHDOG_CHECK_ENABLED=true
...
 ### It has to be a positive integer. Defaults to 24 (hours).
-# GREENBOOT_WATCHDOG_GRACE_PERIOD=24
\ No newline at end of file
+# GREENBOOT_WATCHDOG_GRACE_PERIOD=24
+
+### This variable allows you to specify healthchecks to be skipped. 
+### The healthcheck must be specified by script name, as in the 
+### example below. Multiple healthchecks may be skipped by separating
+### the script names with spaces.
+### DISABLED_HEALTHCHECKS=("01_repository_dns_check.sh 02_watchdog.sh")
```
mmartinv, line 21: Shouldn't this be `DISABLED_HEALTHCHECKS=('01_repository_dns_check.sh' '02_watchdog.sh')`?

usr/libexec/greenboot/greenboot
```diff
@@ -7,6 +7,53 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+    if test -f "$GREENBOOT_CONFIGURATION_FILE"; then
+        source $GREENBOOT_CONFIGURATION_FILE
+    fi
+}
+
+source_configuration_file
+function exists_in_list {
+    VALUE=$1
+    echo "${DISABLED_HEALTHCHECKS,,}" | tr " " '\n' | grep -F -q -x "$VALUE"
+}
```
mmartinv, line 21:
```suggestion
function is_disabled {
    HEALTH_CHECK=$1    
    for DISABLED_HEALTH_CHECK in "${DISABLED_HEALTH_CHECKS[@]}"; do
        [ "${HEALTH_CHECK}" != "${DISABLED_HEALTH_CHECK}" ] || return true
    done
    return false     
}
```
You don't need to rely on external commands to iterate an array

usr/libexec/greenboot/greenboot
```diff
@@ -7,6 +7,53 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+
+script_runner () {
+  local scripts_dir=$1; shift
+  local mode=$1; shift
+  local start_msg=$1; shift
+  local required_hc_failed=false
+  echo "$start_msg"
+  for script in $(find "$scripts_dir" -name '*.sh' | sort); do
+    if exists_in_list "$(basename "$script")"; then
```
mmartinv, line 30:
```suggestion
    if is_disabled "$(basename "$script")"; then
```

usr/libexec/greenboot/greenboot
```diff
@@ -7,6 +7,53 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+    fi
+}
+
+source_configuration_file
+function exists_in_list {
+    VALUE=$1
+    echo "${DISABLED_HEALTHCHECKS,,}" | tr " " '\n' | grep -F -q -x "$VALUE"
+}
+
+script_runner () {
```
mmartinv, line 23: isn't this function also defined at the bottom of this patch?

mmartinv (changes requested):

etc/greenboot/greenboot.conf
djach7: The way I have it currently written I believe it's correct, it takes in a string with multiple healthchecks and then uses spaces to delimit that string. If it's agreed that your method makes more sense then I can work to change it.

usr/libexec/greenboot/greenboot
djach7: The `script_runner` function? I didn't make any changes to where the `script_runner` function was defined so I'm not sure I follow this comment

usr/libexec/greenboot/greenboot
djach7: Making this change for my next commit/force push, your name definitely makes a lot more sense than mine. I'll resolve this convo when I get that commit up

etc/greenboot/greenboot.conf
mmartinv: You are using an array actually, because the string is between parentheses. So you need to use one of 
```bash
DISABLED_HEALTHCHECKS="01_repository_dns_check.sh 02_watchdog.sh"
```
or
```bash
DISABLED_HEALTHCHECKS=("01_repository_dns_check.sh"  "02_watchdog.sh")
``` 

I personally prefer the second approach because you don't execute any external commands, the array can be written in multiple lines and because using files with spaces works as expected:
```bash
DISABLED_HEALTH_CHECKS=(
  "one"
  "TWO"
  "THREE"
  "this is a script with spaces"
);

for DISABLED_HEALTH_CHECK in "${DISABLED_HEALTH_CHECKS[@]}"; do
    echo $DISABLED_HEALTH_CHECK ;
done
```

usr/libexec/greenboot/greenboot
mmartinv: this is one `script_runner` function

usr/libexec/greenboot/greenboot
```diff
@@ -7,6 +8,54 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+      fi
+    fi
+  done
+
+  if [[ $required_hc_failed == true ]]; then
+    exit 1
+  fi
+}
+
 script_runner () {
```
mmartinv, line 59: Here is the other script_runner function, maybe that's why shellcheck is complaining about the non reachable code?

usr/libexec/greenboot/greenboot
```diff
@@ -7,6 +8,54 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+}
+
+script_runner () {
+  local scripts_dir=$1; shift
+  local mode=$1; shift
+  local start_msg=$1; shift
+  local required_hc_failed=false
+  echo "$start_msg"
+  for script in $(find "$scripts_dir" -name '*.sh' | sort); do
+    if is_disabled "$(basename "$script")"; then
```
mmartinv, line 32: I think you need to escape the double quotes surrounding `$script`:
```bash
"$(basename \"$script\")"
```

usr/libexec/greenboot/greenboot
djach7: Found the second one, I was just oblivious on that. I removed the redundancy

usr/libexec/greenboot/greenboot
djach7: I just tested with this, it prevented the healthcheck from being skipped even if specified in the config. Going to leave as is for now

usr/libexec/greenboot/greenboot
djach7: That's fair, I liked the `tr` and `grep` solution for its compactness. I agree that the looping solution you provided is easier to follow, though. (The `grep` solution also made me feel smart for all the fancy flags, but don't tell anyone that.)

usr/libexec/greenboot/greenboot
mmartinv: I see a lot of executions with `"$(basename "$script")"` . Why not using a variable to save the value at the begining of the loop and then use that variable?

Even better, why don't print only the script base name in `find` and use the `${script}` variable directly:
```bash
for script in $(find "${scripts_dir}" -name '*.sh' -type f -printf '%f '); do
    if is_disabled "$script"; then
        echo "'${script}' was skipped, as specified in config"
   else
        local rc=0
        systemd-cat -t "${scripts_dir}/${script}" bash "${script}" || rc=$?
...
```

usr/libexec/greenboot/greenboot
```diff
@@ -7,32 +7,52 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+      else
+        echo "<6>Script '$(basename "$script")' SUCCESS"
+      fi
     fi
   done
 
-  [[ $required_hc_failed == false ]]
+  if [[ $required_hc_failed == true ]]; then
+    exit 1
+  fi
```
mmartinv, line 55: This is almost equivalent but instead of returning an error the function exits the entire program with error, why do you need to change this behavior?

miabbott:
I think this change also needs some validation that the script name(s) specified in `DISABLED_HEALTHCHECKS` actually exist in the different directories that are searched.

Some psuedo-valid bash code:

```bash
DISABLED_HEALTHCHECKS=("01_repository_dns_check.sh 02_watchdog.sh 03_this_script_doesnt_exist.sh")

SCRIPTS_DIRS=("/usr/lib/greenboot/check/required.d"
              "/usr/lib/greenboot/check/wanted.d"
              "/usr/lib/greenboot/green.d"
              "/usr/lib/greenboot/red.d"
              "/etc/greenboot/check/required.d"
              "/etc/greenboot/check/wanted.d"
              "/etc/greenboot/green.d"
              "/etc/greenboot/red.d")

for DISABLED_HEALTH_CHECK in "${DISABLED_HEALTH_CHECKS[@]}"; do
    for SCRIPT_DIR in "${SCRIPTS_DIRS[@]}"; do
        if [ -f "${SCRIPT_DIR}/${DISABLED_HEALTH_CHECK}" ]; then
            echo "DEBUG: Found script; will disable it"
            break
        else
            echo "WARNING: Could not find script named ${DISABLED_HEALTH_CHECK} in any expected location"
        fi
    done
    <do other things with disabling the script>
done
```

mmartinv:
> I think this change also needs some validation that the script name(s) specified in `DISABLED_HEALTHCHECKS` actually exist in the different directories that are searched.
> 
> Some psuedo-valid bash code:
> 
> ```shell
> DISABLED_HEALTHCHECKS=("01_repository_dns_check.sh 02_watchdog.sh 03_this_script_doesnt_exist.sh")
> 
> SCRIPTS_DIRS=("/usr/lib/greenboot/check/required.d"
>               "/usr/lib/greenboot/check/wanted.d"
>               "/usr/lib/greenboot/green.d"
>               "/usr/lib/greenboot/red.d"
>               "/etc/greenboot/check/required.d"
>               "/etc/greenboot/check/wanted.d"
>               "/etc/greenboot/green.d"
>               "/etc/greenboot/red.d")
> 
> for DISABLED_HEALTH_CHECK in "${DISABLED_HEALTH_CHECKS[@]}"; do
>     for SCRIPT_DIR in "${SCRIPTS_DIRS[@]}"; do
>         if [ -f "${SCRIPT_DIR}/${DISABLED_HEALTH_CHECK}" ]; then
>             echo "DEBUG: Found script; will disable it"
>             break
>         else
>             echo "WARNING: Could not find script named ${DISABLED_HEALTH_CHECK} in any expected location"
>         fi
>     done
>     <do other things with disabling the script>
> done
> ```

Actually I don't think it's needed because the function is called with existing scripts only

miabbott:
> Actually I don't think it's needed because the function is called with existing scripts only

Smart!

miabbott:
> Actually I don't think it's needed because the function is called with existing scripts only

I still think there is a chance of a typo when configuring `DISABLED_HEALTHCHECKS` that we should be able to warn users about, right?

Even if the function is called on an existing script (i.e. `01_repository_dns_check.sh`), if I have defined `DISABLED_HEALTHCHECKS=("01_repostory_dns_checks.sh")`, that script will still run because I used the wrong script name.

etc/greenboot/greenboot.conf
say-paul, line 21: I would prefer some something simpler as user is going to modify them.

like: ``` ### DISABLED_HEALTHCHECKS=01_repository_dns_check.sh,02_watchdog.sh```

usr/libexec/greenboot/greenboot
djach7: Rewriting this to be more like what you have above and I just want to make sure I'm understanding it. In what you have above `HEALTHCHECK` does not equal `DISABLED_HEALTHCHECK` returns true, which in the context of the function means "yes it is disabled"? Shouldn't this be reversed, as `HEALTHCHECK` not equaling `DISABLED_HEALTHCHECK` would mean that it isn't in the list, and therefore not disabled? I think I am just confusing myself now.

usr/libexec/greenboot/greenboot
djach7: This was not an intentional change, and I'm not sure why it did. It seems that when if I fail a reboot after making changes and unlocking with `ostree admin unlock --hotfix` that change is already present, even though I never wrote it. Investigating more when I get a chance

usr/libexec/greenboot/greenboot
mmartinv: The `||` operator in bash works this way: 
- It runs the command on the left side and if the returned code was `0` (the command finished successfully) then it skips the command on the right side and continues the execution (in this case it will continue with the next loop). 
- In case the return code is different than `0` (the command failed for whatever reason) then it runs the command on the right side.

So as long as `HEALTHCHECK` is different for the `DISABLED_HEALTHCHECK` the comparison (`[ "${HEALTHCHECK}" != "${DISABLED_HEALTHCHECK}" ]`) will be `true` so the command on the right side will be ignored and the for loop will keep running (we ignore the disabled health checks which are not equal to the healthcheck we are looking for).

On the other hand, if `HEALTHCHECK`  is equal to `DISABLED_HEALTHECK` (we found that this health check is disabled) then the comparison will be `false`, the command on the right side is executed and the `is_disabled` function will return `true`. 

In case the `for` loop ends, it means we didn't found the `HEALTCHECK` on the list (so it is not disabled) and the `is_disabled` function returns `false`.

djach7: Thank you for the explanation!! I have never used that operator in bash before, I thought it was a shorthand if statement (which I guess it sort of is, just not how I was picturing it). I rewrote it to be a one line if statement because that made more sense to me, but if I get a chance I'll try to rework it with the `||`. Thanks again!

etc/greenboot/greenboot.conf
djach7: Currently rewriting to use the second approach you mentioned here, I agree that is probably a simpler method. It will be included in the next force push

djach7: Currently rewriting to use just an array, would that work? Currently looking something like this:

`DISABLED_HEALTHCHECKS=(script1_name script2_name)`

Is this more in line with what you have in mind?

djach7: Addressed in the most recent updates, resolving this issue

usr/libexec/greenboot/greenboot
djach7: Addressed in the most recent updates, I was having issues getting the `||` to work though so I stuck with the if statement. Leaving this open so I remember to keep finagling with the `||`

usr/libexec/greenboot/greenboot
djach7: I tried to implement this in the most recent updates but it kept failing to boot with these changes. I'm sure it's something I'm doing wrong, I just haven't figured it out yet

usr/libexec/greenboot/greenboot
djach7: Addressed and reverted in the most recent changes. Resolving this conversation

djach7:
@miabbott I added a line in the config comments about needing exact spelling. If you'd like me to go beyond that and write something more please let me know

usr/libexec/greenboot/greenboot
```diff
@@ -7,28 +7,49 @@ SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
...
+
+source_configuration_file
+function is_disabled {
+    HEALTHCHECK=$1
+    for DISABLED_HEALTHCHECK in "${DISABLED_HEALTHCHECKS[@]}"; do
+        if [ "${HEALTHCHECK}" == "${DISABLED_HEALTHCHECK}" ]; then return 0; fi
+    done
+    return 1
+}
+
```
7flying, line 26: nit for the two functions above: local variables should be not in uppercase to differentiate them from global variables

etc/greenboot/greenboot.conf
mmartinv: Nitpick, I think it would be better to double quote the script names for coherence with the other arrays we use in the code. 
ExtraMegaNitpick: I think is more clear to write the comment as a multiline, so instead of:
```bash
### DISABLED_HEALTHCHECKS=(01_repository_dns_check.sh 02_watchdog.sh)
```
Something like:
```bash
### DISABLED_HEALTHCHECKS=(
###     "01_repository_dns_check.sh"
###     "02_watchdog.sh"
### )
```

say-paul: I am starting to the like the format that @mmartinv suggested.
Before fully giving in to the idea, I would think form user perspective - Does this format makes it easy for some config preparer to easily update the configs.

say-paul:
Apart from the config format, the changes looks good to me.

say-paul (approved):

miabbott:
> @miabbott I added a line in the config comments about needing exact spelling. If you'd like me to go beyond that and write something more please let me know

It's a good first step, but I still think we should be able to warn the user if something in `DISABLED_HEALTHCHECKS` can't be found.  I don't want to block further progress on this implementation, so maybe you can do that kind of warning in a follow-up PR.

etc/greenboot/greenboot.conf
djach7: Updated to reflect the format suggested above, resolving this conversation

PR #150, opened 2024-08-22
packit.yaml: add copy_upstream_release_description true

pcdubs:
- Added `copy_upstream_release_description true` to packit.yaml to fix missing downstream change log entries.
- fix date entry in change log
- migrated to SPDX license (dist-git change)


say-paul:
LGTM

PR #151, opened 2024-08-27
Fixed the issue that boot_counter cannot be unset and some scripts do…

ssk-wh:
… not have executable permissions


nullr0ute (approved):
LGTM

PR #152, opened 2024-09-05
fix: remove bootupd

say-paul:
moved greenboot.cfg to /etc/grub.d


miabbott:
Attempting to workaround/fix - https://github.com/fedora-iot/iot-distro/issues/58

greenboot.spec
```diff
@@ -90,6 +90,9 @@ install -DpZm 0644 etc/greenboot/greenboot.conf %{buildroot}%{_sysconfdir}/%{nam
 %systemd_post greenboot-grub2-set-success.service
 %systemd_post greenboot-rpm-ostree-grub2-check-fallback.service
 %systemd_post redboot-auto-reboot.service
+if [ -d /usr/lib/bootupd/grub2-static/configs.d ]; then
```
nullr0ute, line 93: Nope! This is not going to work.

greenboot.spec
```diff
@@ -142,7 +145,7 @@ install -DpZm 0644 etc/greenboot/greenboot.conf %{buildroot}%{_sysconfdir}/%{nam
 %dir %{_prefix}/lib/%{name}/red.d
 %{_exec_prefix}/lib/motd.d/boot-status
 %{_tmpfilesdir}/greenboot-status-motd.conf
-%{_prefix}/etc/grub.d/*.cfg
+%{_sysconfdir}/grub.d/*.cfg
```
nullr0ute, line 148: This is an unrelated fix, it should be in a different commit/PR.

greenboot.spec
miabbott: @nullr0ute do you have any suggestions for conditionalizing where the grub config lands as Sayan is trying to do?

nullr0ute: The general rule of thumb is that you shouldn't do it. I think we need to step back and look at the actual problem.

greenboot.spec
say-paul: I think you saw comparison between two commits, which made it seem unrelated.
Its actually changing the path from bootupd/static-config to grub.d

greenboot.spec
say-paul: So I applied a hard link instead of mv also added in %postun to remove the ln post installation.

ggiguash: What about creating a symlink to the file instead of moving it? 
This way we can also undo the symlink creation in the %postun section

ggiguash: Is there a chance that /etc would be mounted on a different flesystem? 
Is symlink better for what we're trying to do?

say-paul: Thats a good point, hardlink will have issue between different file systems, softlink didn't work so I plan on using `cp` as it may not have an integrity issue as pointed out by @nullr0ute

Issue #157, opened 2024-09-25
Greenboot produces a warning message when DISABLED_HEALTHCHECKS is used


dhensel-rh:
Testing `greenboot-0.15.6-1.el9_4.x86_64` 

When I use the DISABLED_HEALTHCHECKS variable in /etc/greenboot/greenboot.conf, I am receiving a Warning message

Sep 25 15:24:14 microshift-dev.local 50_microshift_running_check_olm.sh[5299]: FINISHED
Sep 25 15:24:14 microshift-dev.local greenboot[851]: Script '50_microshift_running_check_olm.sh' SUCCESS
Sep 25 15:24:14 microshift-dev.local greenboot[851]: Running Wanted Health Check Scripts...
Sep 25 15:24:14 microshift-dev.local greenboot[851]: WARNING: 01_repository_dns_check.sh was not found and may be misspelled

This seems unexpected since I asked that `01_repository_dns_check.sh` be skipped.

Contents of /etc/greenboot/greenboot.conf

```
### This variable allows you to specify healthchecks to be skipped.
### The healthcheck must be specified by script name, as in the
### example below. Multiple healthchecks may be skipped by separating
### the script names with spaces.
### NOTE: Script names must be spelled EXACTLY. Typos will result in
### unwanted behaviour.
### DISABLED_HEALTHCHECKS=(
###     "01_repository_dns_check.sh"
###     "02_watchdog.sh"
### )
DISABLED_HEALTHCHECKS=("01_repository_dns_check.sh")

```

PR #160, opened 2024-10-22
fix: reword warning message for disabled checks

runcom:
Fix https://github.com/fedora-iot/greenboot/issues/157

Just changes the warning message to clarify why the check hasn't run.


nullr0ute (approved):
Looks much clearer

PR #193, opened 2024-12-05
ci: fix rhel 9.6 anaconda case failure

yih-redhat:


nullr0ute:
Generally the point of pull requests is for review so it's generally considered not the right thing to merge it yourself, especially with about 30 mins of opening the PR.

PR #197, opened 2025-01-15
specfile: Stop building on ix86

travier:
Initially pushed downstream in Fedora [1], but overwrote by packit with the 0.15.7 update [2], thus pushed here.

[1]: https://src.fedoraproject.org/rpms/greenboot/pull-request/30
[2]: https://src.fedoraproject.org/rpms/greenboot/pull-request/35

See: https://fedoraproject.org/wiki/Changes/EncourageI686LeafRemoval


miabbott (approved):
Thanks for catching this, Timothee!

LGTM

Issue #143, opened 2024-07-10
Greenboot does not fail when first error is found


dhensel-rh:
This issue is observed in MicroShift project. When multiple start scripts are on the system, and an error is encountered in the first script, Greenboot does not immediately fail. Using default settings, Greenboot checks all of them before rebooting (in 4.16, this is 3 scripts 40_microshift_running_check.sh, 41_microshift_running_check_multus.sh, 50_microshift_running_check_olm.sh). This can add delay to the system rolling back to a known good state. This might be intended design, but it can add time before a rollback can occur.

On the first boot, each script can take up to 5 minutes to check. That is ~15 minutes. On the second boot, this increases to 10 minutes per script. That is ~30 minutes. On the third boot, each script takes up to 15 minutes. 

Can this be optimized in some way ?

nullr0ute:
I think the microshift scripts are part of microshift, the bug would need to be reported there as those are the scripts that are taking the time

dhensel-rh:
I opened a ticket wit hMicroShift and discussed this with Gregory  Giguashvili previously. He felt this was a Greenboot feature enhancement.

'Looking at https://github.com/fedora-iot/greenboot/blob/main/usr/libexec/greenboot/greenboot#L69, the greenboot check continues the loop until all scripts are run and only checks for the error condition at https://github.com/fedora-iot/greenboot/blob/main/usr/libexec/greenboot/greenboot#L78.`

nullr0ute:
Can you provide a link to the microshift ticket, the details provided to date don't provide that context. Also links to where the mentioned scripts can be found.

dhensel-rh:
Oh yes. i was not sure if you could view the ticket. Apologies

https://issues.redhat.com/browse/USHIFT-3165

[path to MicroShift greenboot files](https://github.com/openshift/microshift/tree/27a2e26e71b22ef8da614971c451924cc996fbf6/packaging/greenboot)

nullr0ute:
Seems like it should be a RFE to have an option to either 1) run all tests 2) fail at the first failure. To quote the full comment:

`
[Douglas Hensel](https://issues.redhat.com/secure/ViewProfile.jspa?name=rh-ee-dhensel) , this is by design of the current greenboot implementation.

[...]

PR #198, opened 2025-01-28
fix:fail early for any required script failure

say-paul:
When greenboot encounters an error in required scripts, it breaks out the healtcheck loop and moves to red script execution. Logs are added to notify users of the skipped scripts due to critical failure.

fixes: #143

### DIFF
--- a/usr/libexec/greenboot/greenboot
+++ b/usr/libexec/greenboot/greenboot
@@ -6,7 +6,6 @@ LC_ALL=C
 SCRIPTS_CHECK_PATHS=("/usr/lib/greenboot/check" "/etc/greenboot/check")
 SCRIPTS_GREEN_PATHS=("/usr/lib/greenboot/green.d" "/etc/greenboot/green.d")
 SCRIPTS_RED_PATHS=("/usr/lib/greenboot/red.d" "/etc/greenboot/red.d")
-declare -A DIS_CHECK_STATUS=()
 
 source_configuration_file() {
     greenboot_configuration_file=/etc/greenboot/greenboot.conf
@@ -17,34 +16,15 @@ source_configuration_file() {
 }
 
 source_configuration_file
-function init_disabled_map {
-    for disabled_healthcheck in "${DISABLED_HEALTHCHECKS[@]}"; do
-	DIS_CHECK_STATUS["$disabled_healthcheck"]=1;
-    done
-}
-
-source_configuration_file
-function print_unexecuted_checks {
-    for disabled_healthcheck in "${DISABLED_HEALTHCHECKS[@]}"; do
-	if [[  "${DIS_CHECK_STATUS[$disabled_healthcheck]}" == 1 ]]; then
-           echo "WARNING: $disabled_healthcheck wasn't executed because it's disabled"
-	fi
-    done
-}
-
-source_configuration_file
 function is_disabled {
     healthcheck=$1
     for disabled_healthcheck in "${DISABLED_HEALTHCHECKS[@]}"; do
-        if [ "${healthcheck}" == "${disabled_healthcheck}" ]; then
-	    DIS_CHECK_STATUS["${disabled_healthcheck}"]=0
+      if [ "${healthcheck}" == "${disabled_healthcheck}" ]; then
 	    return 0
 	fi
     done
     return 1
 }
-
-init_disabled_map
 
 script_runner () {
   local scripts_dir=$1; shift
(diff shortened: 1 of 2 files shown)


usr/libexec/greenboot/greenboot
```diff
@@ -82,7 +84,10 @@ case "$1" in
...
     rc=0
     for health_check_path in "${SCRIPTS_CHECK_PATHS[@]}"; do
-      script_runner "$health_check_path/required.d" "strict" "Running Required Health Check Scripts..." || rc=1
+      script_runner "$health_check_path/required.d" "strict" "Running Required Health Check Scripts..." || {
+        rc=1
+        break
+      }
       script_runner "$health_check_path/wanted.d" "relaxed" "Running Wanted Health Check Scripts..."
     done
     print_unexecuted_checks
```
say-paul, line 93: @djach7  do you remember why printing the disabled scripts again when it's already printed in script_runner()?

runcom (commented):
LGTM overall, will wait for @djach7 ACK to clear up @say-paul 's comment 
although, tests are failing somehow

usr/libexec/greenboot/greenboot
djach7: I don't remember, no. That doesn't make much sense to me reading it now.

say-paul: remove scripts that are irrelevant.

LorbusChris:
I just wanted to note that having all tests run and not breaking after the first failure is actually a feature, not a bug, as in case of multiple failures each one will be reported (otherwise following failures will only be uncovered when the first one is fixed and the device is booted again). Maybe that isn't needed, but OTOH this could be made optional with e.g. an `--exit-immediately` flag

usr/libexec/greenboot/greenboot
```diff
@@ -16,45 +15,28 @@ source_configuration_file() {
...
   local start_msg=$1; shift
   local required_hc_failed=false
   echo "$start_msg"
   for script in $(find "$scripts_dir" -name '*.sh' | sort); do
-    if is_disabled "$(basename "$script")"; then
-      echo "'$(basename "$script")' was skipped, as specified in config"
+    if [[ "$required_hc_failed" == true && "$mode" == "strict" ]]; then
+      echo "<5>'$(basename "$script")' was skipped due to a previous failure"
+    elif is_disabled "$(basename "$script")"; then
+      echo "<5>'$(basename "$script")' was skipped, as specified in config"
```
sarmahaj, line 39: can we pls update warning /error msg on failure on line 48 & 52.
 "relaxed")
            echo "<2>$failure_msg. Continuing..." >&2  -> continuing since relaxed mode or something
        and for 
 "strict")
            required_hc_failed=true
            echo "<0>$failure_msg. Continuing..." >&2 -> Exiting due to required script failure

usr/libexec/greenboot/greenboot
```diff
@@ -16,45 +15,28 @@ source_configuration_file() {
...
     else
       local rc=0
```
sarmahaj, line 41: can this `rc` variable be renamed to something providing clarity what is it meant for , like is it a return code or some count ?

usr/libexec/greenboot/greenboot
```diff
@@ -82,10 +64,12 @@ case "$1" in
   "check")
     rc=0
     for health_check_path in "${SCRIPTS_CHECK_PATHS[@]}"; do
-      script_runner "$health_check_path/required.d" "strict" "Running Required Health Check Scripts..." || rc=1
+      script_runner "$health_check_path/required.d" "strict" "Running Required Health Check Scripts..." || {
```
sarmahaj, line 67: I understand the logic part here but why do we need OR (||) operator here?  
can we not just use something like
If return val of script_runner !=0 i.e failure then break .
(or is it the bash thing !?)

runcom:
> I just wanted to note that having all tests run and not breaking after the first failure is actually a feature, not a bug, as in case of multiple failures each one will be reported (otherwise following failures will only be uncovered when the first one is fixed and the device is booted again). Maybe that isn't needed, but OTOH this could be made optional with e.g. an `--exit-immediately` flag

right, this could be enhanced for sure to have something that specifies that we want to exit as soon as a failure happens. Although if even one of the healthcheck fails, we're going to rollback anyway and surely at the next upgrade something else may fail but that's the flow anyway right?

usr/libexec/greenboot/greenboot
say-paul: This is internal to greenboot as when strict/relaxed is used , it logs for users in terms of required/wanted/green/red scripts adding relax/strict will add to the confusion.

usr/libexec/greenboot/greenboot
say-paul: both approaches yields same result, I don't see what `if..then ..` offers more?

runcom:
@mcattamoredhat how was testing for this? Good to go?

PR #199, opened 2025-02-06
rollback: if no bootc, fallback to rpm-ostree

runcom:
Newer rpm-ostree in 9.6 pulls in bootc, which made our test fails because we weren't rolling back the deployment.
This patch puts the `return` where it should be in case we're not "bootc-native". That way we can safely fallback to rpm-ostree rollback (bootc rollback should work tho...). A fallout of this nasty bug is that we don't get any logs either so just adding some for mental sanity.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
 
       - if: ${{ always() }}
         name: Upload artifact with ShellCheck defects in SARIF format
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


runcom:
@yih-redhat as a follow up to this PR, we need to have a test that does upgrade from 9.5 to 9.6 and makes sure the rollback works - it's probably best if we have that test in osbuild-composer

say-paul:
LGTM

runcom:
do not merge until we have a sanity check from @mcattamoredhat in https://github.com/osbuild/osbuild-composer/issues/4593

runcom:
Tests from Mario went green, so merged

PR #203, opened 2025-02-19
packit: sync with fdo to have propose downstream in centos too

runcom:
As the title says, just sync packit config with https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/.packit.yaml


mmartinv (approved):
LGTM

PR #205, opened 2025-02-27
do not build ix86 on rhel<10 and fedora only

runcom:


miabbott:
LGTM

runcom:
as noted by @yih-redhat these failures aren't greenboot related, getting this as it's just aspec changes

PR #214, opened 2025-03-18
Bootupd grub2 static ordering

### LINKED ISSUE #213: /etc/grub.d/greenboot.cfg always ignored
https://github.com/rhboot/grub2/blame/569214e39d89a0fe776131ac613b69b498901078/util/grub-mkconfig.in#L305
grub-mkconfig expects an executable in /etc/grub.d/ else it just skips the file
greenboot.cfg was added for bootupd, not for grub.d

Bug introduced in https://github.com/fedora-iot/greenboot/commit/2b36530190f77d7b128a87d18335269e56797437
The proper fix was to just remove `Requires: bootupd`

champtar:
Fixes #213 
Needed for https://github.com/coreos/bootupd/pull/841


cgwalters:
> The only fix needed was to drop 'Requires: bootupd'

Your commit message says this but I don't see it? Not totally sure if that'd be right anyways?

champtar:
> > The only fix needed was to drop 'Requires: bootupd'
> 
> Your commit message says this but I don't see it?

The commit I'm partially reverting was removing `Requires: bootupd`, so I didn't re-add it to re-remove it
I can do a full revert if prefered

> Not totally sure if that'd be right anyways?

What problem are you thinking about ?
If you don't like not having the dependency then we need a separate package
The hack of moving a file in `%post` depend on the install order, and prevent the use of `dnf provides`

cgwalters:
This is all kind of head-spinning and inherently fragile, but AIUI at least this project has a history of wanting to support non-bootupd methods of operation because basically when Fedora IoT forked from CoreOS (ish) at the time CoreOS wasn't using bootupd.

We have *years* of technical debt to unwind here.

champtar:
I'm not changing any dependencies here, just putting the file directly in `/usr/lib/bootupd/grub2-static/configs.d`, if bootupd is not present nothing happens, that's it

say-paul:
we have kept it that way to keep Greenboot backward compatible with non-bootupd OS. I suspect the changes introduced will affect rpm-ostree systems.

champtar:
> we have kept it that way to keep Greenboot backward compatible with non-bootupd OS. I suspect these changes will affect rpm-ostree systems.

As written in my commit message:
```
/etc/grub.d expects scripts that print the config on stdout
/usr/lib/bootupd/grub2-static/configs.d expects config
```
/etc/grub.d/greenboot.cfg is ignored, it was never meant to be used in /etc/grub.d/, and greenboot.cfg is the output of `/etc/grub.d/08_fallback_counting` plus part of `/etc/grub.d/10_reset_boot_success` which are provided by grub2-tools

These changes make the bootupd ordering match the legacy grub ordering,
they change nothing for non bootupd system.

yih-redhat:
@say-paul rpm-ostree image test has passed on RHEL9.6 and Centos-stream-9

say-paul (approved):

champtar:
> We will do a release here to trigger rhe process.

Perfect ! Any approximate time frame ? (this is the last blocker for https://github.com/coreos/bootupd/pull/841)

PR #229, opened 2026-02-10
Chore(deprecation): Deprecating greenboot in favour of greenboot-rs

say-paul:
development in rust moving to: https://github.com/fedora-iot/greenboot-rs.git


miabbott (approved):
LGTM

